PR #3, opened 2019-03-27
Add support for path in service role

mcieplak:
@sureshvis

### DIFF
--- a/pkg/servicerole/servicerole.go
+++ b/pkg/servicerole/servicerole.go
@@ -18,6 +18,8 @@ import (
 	"istio.io/istio/pilot/pkg/model"
 )
 
+const emptyPath = "empty-path"
+
 var client *crd.Client
 
 func init() {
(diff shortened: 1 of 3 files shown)


pkg/servicerole/servicerole.go
```diff
@@ -61,9 +63,19 @@ func createServiceRole(namespace, role string, policy *zms.Policy) (model.Config
 		Namespace: namespace,
 	}
 
-	methods := make([]string, 0)
+	pathToMethods := make(map[string][]string)
 	for _, assertion := range policy.Assertions {
-		methods = append(methods, strings.ToUpper(assertion.Action))
+		key := emptyPath
+		path := strings.Split(assertion.Resource, ":")
```
sureshvis: comment with example how its defined on athen policy section

mcieplak: Added the comment.

PR #13, opened 2019-06-12
Add ClusterRbacConfig controller, work queue, and full sync

mcieplak:
@prabushyam


pkg/istio/onboarding/controller.go
```diff
@@ -0,0 +1,246 @@
...
+		return c.createClusterRbacConfig(serviceList)
+	}
+
+	clusterRbacConfig, ok := config.Spec.(*v1alpha1.RbacConfig)
+	if !ok {
+		return errors.New("Could not cast to ClusterRbacConfig")
+	}
+
+	if clusterRbacConfig.Inclusion == nil {
+		return errors.New("ClusterRbacConfig inclusion service list is empty")
```
prabushyam: is this an error condition, in the case where the crc resource exists, but not the `Inclusion` field, this would return an error... if there are services to be onboarded, we could add the inclusion field and then process the service list

pkg/istio/onboarding/controller.go
```diff
@@ -0,0 +1,246 @@
...
+			Spec:       clusterRbacConfig,
+		})
+		return err
+	}
+
+	log.Println("Sync state is current, no changes needed...")
+	return nil
+}
+
+// compareServices compares the two arrays and returns the difference
```
prabushyam: can we update the comment to explicitly indicate that this func returns the items in first list that are not in the second list

pkg/istio/onboarding/controller.go
```diff
@@ -0,0 +1,246 @@
...
+		if _, exists := serviceMapB[item]; !exists {
+			serviceListDiff = append(serviceListDiff, item)
+		}
+	}
+
+	return serviceListDiff
+}
+
+// remove removes an element from an array at the given index
+func removeIndexElement(serviceList []string, indexToRemove int) []string {
```
prabushyam: 💯  nice optimization

pkg/istio/onboarding/controller_test.go
```diff
@@ -0,0 +1,401 @@
...
+			name:             "test processing a service object",
+			inputObject:      onboardedService,
+			expectedQueueLen: 1,
+			expectedKeyName:  "test-namespace/onboarded-service",
+		},
+		{
+			name:             "test processing a non service object",
+			inputObject:      nil,
+			expectedQueueLen: 0,
+		},
```
prabushyam: could we add a notOnboardedService test case being added to queue as well?

pkg/istio/onboarding/controller_test.go
```diff
@@ -0,0 +1,401 @@
...
+func TestAddService(t *testing.T) {
+	c := &Controller{}
+
+	tests := []struct {
+		name              string
+		inputServices     []string
+		clusterRbacConfig *v1alpha1.RbacConfig
+		expectedArray     []string
+	}{
+		{
```
prabushyam: could we add a nil `clusterRbacConfig:` test case?

prabushyam: also, an empty `inputServices` list?

pkg/istio/onboarding/controller_test.go
```diff
@@ -0,0 +1,401 @@
...
+			name:              "test deleting service from ClusterRbacConfig",
+			clusterRbacConfig: createClusterRbacConfig(onboardedServiceName),
+			inputArray:        []string{onboardedServiceName},
+			expectedArray:     []string{},
+		},
+		{
+			name:              "test deleting existing service from ClusterRbacConfig",
+			clusterRbacConfig: createClusterRbacConfig(onboardedServiceName, existingServiceName),
+			inputArray:        []string{onboardedServiceName},
+			expectedArray:     []string{existingServiceName},
```
prabushyam: could we add test cases for empty inputArray, and a servce that doesn't exist on the clusterRbacConfig

pkg/istio/onboarding/controller.go
```diff
@@ -0,0 +1,246 @@
...
+)
+
+type Controller struct {
+	store                model.ConfigStoreCache
+	dnsSuffix            string
+	serviceIndexInformer cache.SharedIndexInformer
+	queue                workqueue.RateLimitingInterface
+}
+
+type metaNamespaceKeyFunc func(obj interface{}) (string, error)
```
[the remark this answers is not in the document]
mcieplak: Ah nice! Updated it.

pkg/istio/onboarding/controller.go
```diff
@@ -0,0 +1,246 @@
...
+		dnsSuffix:            dnsSuffix,
+		serviceIndexInformer: serviceIndexInformer,
+		queue:                queue,
+	}
+
+	serviceIndexInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			c.processEvent(cache.MetaNamespaceKeyFunc, obj)
+		},
+		UpdateFunc: func(old interface{}, new interface{}) {
```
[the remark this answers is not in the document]
mcieplak: New surprisingly isn't a keyword but new() is a function which is supported which initializes an object with empty values. I have renamed it to avoid confusion.

pkg/istio/onboarding/controller.go
```diff
@@ -0,0 +1,246 @@
...
+	key, quit := c.queue.Get()
+	if quit {
+		return false
+	}
+
+	defer c.queue.Done(key)
+
+	err := c.sync()
+	if err != nil {
+		log.Println("Error syncing cluster rbac config:", err)
```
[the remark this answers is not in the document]
mcieplak: Yep, updated!

pkg/istio/onboarding/controller.go
```diff
@@ -0,0 +1,246 @@
...
+		if c.queue.NumRequeues(key) < numRetries {
+			c.queue.AddRateLimited(key)
```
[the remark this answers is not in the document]
mcieplak: Sure, added it.

pkg/istio/onboarding/controller.go
```diff
@@ -0,0 +1,246 @@
...
+	}
+
+	c.queue.Forget(key)
+	return true
+}
+
+// addService will add a service to the ClusterRbacConfig object
+func (c *Controller) addServices(services []string, clusterRbacConfig *v1alpha1.RbacConfig) {
+	for _, service := range services {
+		clusterRbacConfig.Inclusion.Services = append(clusterRbacConfig.Inclusion.Services, service)
```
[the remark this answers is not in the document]
mcieplak: Yes, good point. Since we are removing the check from the sync function, I have added it here now.

pkg/istio/onboarding/controller.go
```diff
@@ -0,0 +1,246 @@
...
+		},
+	}
+
+	_, err := c.store.Create(config)
+	return err
+}
+
+// getServiceList extracts all services from the indexer with the authz
+// annotation set to true.
+func (c *Controller) getServiceList() []string {
```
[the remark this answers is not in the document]
mcieplak: Renamed to getOnboardedServiceList.

pkg/istio/onboarding/controller.go
mcieplak: That's a good point, it could still exist and we should add service to the object in those cases. I have removed this if statement for those use cases.

pkg/istio/onboarding/controller.go
mcieplak: Updated!

pkg/istio/onboarding/controller.go
mcieplak: Thanks to stack overflow :)

pkg/istio/onboarding/controller_test.go
mcieplak: Added the test case.

pkg/istio/onboarding/controller_test.go
mcieplak: Sure, added the tests.

pkg/istio/onboarding/controller_test.go
mcieplak: Updated.

mcieplak:
@prabushyam I have updated the PR, please review again. Thanks!

PR #14, opened 2019-06-12
 Implement RBAC changes 

prabushyam:
@mcieplak


pkg/istio/servicerole/servicerole.go
```diff
@@ -53,14 +75,151 @@ func (srMgr *ServiceRoleMgr) GetServiceRoleMap() (map[string]*ServiceRoleInfo, e
...
+
+	rules := make([]*v1alpha1.AccessRule, 0)
+	for _, assertion := range assertions {
+		assertionRole, err := common.ParseRoleFQDN(domainName, string(assertion.Role))
+		if err != nil {
+			log.Println(err.Error())
+			continue
+		}
+
+		if assertionRole != roleName {
```
mcieplak: Shouldn't the assertions already be grouped by the role at this point? Why are we doing this check?

pkg/istio/servicerole/servicerole.go
```diff
@@ -53,14 +75,151 @@ func (srMgr *ServiceRoleMgr) GetServiceRoleMap() (map[string]*ServiceRoleInfo, e
...
+			continue
+		}
+
+		svc, path, err := parseAssertionResource(domainName, assertion)
+		if err != nil {
+			log.Println(err.Error())
+			continue
+		}
+
+		rule := &v1alpha1.AccessRule{}
```
mcieplak: Can we create one object here and set all these values?

pkg/istio/servicerole/servicerole.go
```diff
@@ -53,14 +75,151 @@ func (srMgr *ServiceRoleMgr) GetServiceRoleMap() (map[string]*ServiceRoleInfo, e
...
+		constraint := &v1alpha1.AccessRule_Constraint{
+			Key:    ConstraintSvcKey,
+			Values: []string{svc},
+		}
+		rule.Constraints = []*v1alpha1.AccessRule_Constraint{constraint}
+		rule.Methods = []string{method}
+		if path != "" {
+			rule.Paths = []string{path}
```
mcieplak: Will this allow all paths by default if not specified?

pkg/controller/controller.go
```diff
@@ -89,6 +91,25 @@ func (c *Controller) sync() error {
...
+		signedDomain, err := zms.GetSignedDomain(domainName)
+		if err != nil {
+			log.Println("Error getting the signed domain from ZMS:", err)
+			continue
+		}
+
+		domainRBAC := m.ConvertAthenzPoliciesInToRBACModel(signedDomain.Domain)
+
+		provider := rbacv1.NewProvider()
+		rbacCRs := provider.ConvertAthenzModelIntoIstioRBAC(domainRBAC)
```
mcieplak: I am thinking we need to add the validation functions calls somewhere for the specs.

pkg/istio/servicerole/servicerole_test.go
```diff
@@ -48,3 +50,248 @@ func TestCreateServiceRole(t *testing.T) {
...
+			expectedSpec: &v1alpha1.ServiceRole{
+				Rules: []*v1alpha1.AccessRule{
+					{
+						Methods: []string{
+							"GET",
+						},
+						Paths: []string{
+							"/protected/path",
+						},
+						Constraints: []*v1alpha1.AccessRule_Constraint{
```
mcieplak: Can we add a test case here without a path?

pkg/istio/servicerolebinding/servicerolebinding.go
```diff
@@ -105,6 +108,75 @@ func (srbMgr *ServiceRoleBindingMgr) getSubjects(members []zms.MemberName) []*v1
...
+	return common.PrincipalToSPIFFE(memberStr)
+}
+
+// GetServiceRoleBindingSpec returns the ServiceRoleBindingSpec for a given Athenz role and its members
+func GetServiceRoleBindingSpec(roleName string, members []*zms.RoleMember) (*v1alpha1.ServiceRoleBinding, error) {
+
+	subjects := make([]*v1alpha1.Subject, 0)
+	for _, member := range members {
+
+		//TODO: handle member.Expiration for expired members, for now ignore expiration
```
mcieplak: Does Athenz themselves remove the expired members?

pkg/istio/servicerolebinding/servicerolebinding.go
```diff
@@ -19,7 +21,8 @@ import (
 )
 
 const (
-	allUsers = "user.*"
+	allUsers         = "user.*"
+	wildCardAllUsers = "*"
```
mcieplak: Does istio have a constant for this wildcard use case which we could use?

pkg/istio/servicerolebinding/servicerolebinding_test.go
```diff
@@ -40,3 +43,111 @@ func TestCreateServiceRoleBinding(t *testing.T) {
...
+			expectedErr:    nil,
+		},
+		{
+			test: "valid wildcard member",
+			member: &zms.RoleMember{
+				MemberName: zms.MemberName("user.*"),
+			},
+			expectedMember: "*",
+			expectedErr:    nil,
+		},
```
mcieplak: Can we also add one test case with an invalid member name.

pkg/istio/servicerolebinding/servicerolebinding_test.go
```diff
@@ -40,3 +43,111 @@ func TestCreateServiceRoleBinding(t *testing.T) {
...
+		expectedErr  error
+	}{
+		{
+			test: "empty args",
+			input: input{
+				roleName: "",
+				members:  nil,
+			},
+			expectedSpec: nil,
+			expectedErr:  fmt.Errorf("no subjects found for the ServiceRoleBinding: "),
```
mcieplak: shouldn't this have the role name at the end of the string?

pkg/istio/servicerolebinding/servicerolebinding_test.go
```diff
@@ -40,3 +43,111 @@ func TestCreateServiceRoleBinding(t *testing.T) {
...
+				Subjects: []*v1alpha1.Subject{
+					{
+						User: "athenz.domain/sa/client-serviceA",
+					},
+					{
+						User: "user/sa/athenzuser",
+					},
+				},
+			},
+			expectedErr: nil,
```
mcieplak: Can we also add an invalid member test case?

mcieplak:
@prabushyam Mostly looks good overall, had mainly minor comments. Unfortunately you will have to resolve the merge conflicts as well.

pkg/istio/servicerole/servicerole.go
```diff
@@ -53,14 +75,151 @@ func (srMgr *ServiceRoleMgr) GetServiceRoleMap() (map[string]*ServiceRoleInfo, e
 	return serviceRoleMap, nil
 }
 
+// NewConfig returns a new model.Config resource with the given namespace/name and spec
+func NewConfig(namespace string, name string, spec *v1alpha1.ServiceRole) model.Config {
```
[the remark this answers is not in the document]
prabushyam: will remove the createServiceRole function with a subsequent PR

pkg/istio/servicerolebinding/servicerolebinding.go
```diff
@@ -105,6 +108,75 @@ func (srbMgr *ServiceRoleBindingMgr) getSubjects(members []zms.MemberName) []*v1
 	return subjects
 }
 
+// NewConfig returns a new model.Config resource with the given namespace/name and spec
+func NewConfig(namespace string, name string, spec *v1alpha1.ServiceRoleBinding) model.Config {
```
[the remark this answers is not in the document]
prabushyam: sure, moved to common/util.go

pkg/istio/rbac/common/util.go
```diff
@@ -0,0 +1,35 @@
...
+	"fmt"
+	"github.com/yahoo/athenz/clients/go/zms"
+	"strings"
+)
+
+// ParseRoleFQDN parses the Athenz role full name in the format <domainName>:role.<roleName> to roleName
+// e.g. app-domain:role.reader -> reader
+func ParseRoleFQDN(domainName zms.DomainName, roleFQDN string) (string, error) {
+	roleName := strings.TrimPrefix(roleFQDN, fmt.Sprintf("%s:role.", domainName))
+	if strings.Contains(roleName, ":") {
```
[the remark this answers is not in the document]
prabushyam: yep, trust roles can have another domain prefixed role names, which at the moment we cannot support. We need to think through that use case.

pkg/istio/rbac/common/util.go
```diff
@@ -0,0 +1,35 @@
...
+		return "", fmt.Errorf("role: %s does not belong to the Athenz domain: %s", roleFQDN, domainName)
+	}
+	return roleName, nil
+}
+
+// PrincipalToSPIFFE converts the Athenz principal into a SPIFFE compliant format
+// e.g. client-domain.frontend.some-app -> client-domain.frontend/sa/some-app
+func PrincipalToSPIFFE(principal string) (string, error) {
```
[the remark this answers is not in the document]
prabushyam: For handling the `<domain>:role.` prefix extraction? I couldn't find one in Go...

pkg/istio/rbac/common/util_test.go
```diff
@@ -0,0 +1,93 @@
...
+			domainName:   "athenz.domain",
+			roleFQDN:     "another-domain:role.reader-role",
+			expectedRole: "",
+			expectedErr:  fmt.Errorf("role: another-domain:role.reader-role does not belong to the Athenz domain: athenz.domain"),
+		},
+	}
+
+	for _, c := range cases {
+		gotRole, gotErr := ParseRoleFQDN(c.domainName, c.roleFQDN)
+		a := assert.New(t)
```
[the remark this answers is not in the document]
prabushyam: assert.Equal() takes in `t` , I guess the current approach would re-user the `assert.New(t)` object for all the assertions

pkg/istio/servicerole/servicerole.go
prabushyam: yep, this is extra safety check, since the actual grouping was done by athenz.model, and this one is rbac provider implementation specific, wanted to have the check explicit

pkg/istio/servicerole/servicerole.go
prabushyam: yup, that should be expected behavior

pkg/controller/controller.go
prabushyam: good catch. totally missed the validation. added after generating each SR/SRB spec...

pkg/istio/servicerolebinding/servicerolebinding.go
prabushyam: Couldn't find exportable constant for wildcard, let me know if you found one

pkg/istio/servicerolebinding/servicerolebinding.go
prabushyam: I think ZTS / ZMS does the expiration check when the role members are accessed (during checkAccess call), so ideally we should be doing while we get members of a role during every sync

pkg/istio/servicerolebinding/servicerolebinding_test.go
prabushyam: the passed-in role name is empty, so the error message has nothing to print

prabushyam:
@mcieplak  thank you for the review. I marked the fixed comments as resolved, and left the ones unresolved that I responded in-line. Please re-review when you get a chance

prabushyam:
I added the validation and that caught 2 issues:
- ServiceRole needs to have at least 1 `service` in the rule in 1.1 release, this validation check was removed in 1.2. So, I added `services: *` for all rules automatically. 
- ServiceRoleBinding RoleRef validation looks for `ServiceRole` string instead of `service-role` that the current code was trying to do, fixed both.

pkg/athenz/model_test.go
```diff
@@ -0,0 +1,431 @@
+// Copyright 2019, VerizonMedia Inc.
+// Licensed under the terms of the 3-Clause BSD license. See LICENSE file in github.com/yahoo/k8s-athenz-istio-auth
+// for terms.
+package athenz
+
+import (
+	"github.com/ardielle/ardielle-go/rdl"
+	"github.com/yahoo/athenz/clients/go/zms"
+	"reflect"
+	"testing"
```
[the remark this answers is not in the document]
mcieplak: Can we update the imports in main.go and model_test.go as well?

PR #16, opened 2019-06-13
Add initial workqueue to main controller

mcieplak:
@prabushyam


pkg/controller/controller.go
```diff
@@ -250,12 +260,45 @@ func (c *Controller) Run(stop chan struct{}) {
...
+	if quit {
+		return false
+	}
+
+	defer c.queue.Done(key)
+
+	// TODO, change to new sync function
+	err := c.sync()
+	if err != nil {
+		log.Printf("Error syncing athenz state for key %s: %s", key, err)
```
prabushyam: might need a `err.Error()` for `%s` ?

pkg/controller/controller.go
```diff
@@ -205,6 +213,7 @@ func (c *Controller) sync() error {
 // 7. Namespace informer / indexer
 // 8. Service informer
 func NewController(pollInterval time.Duration, dnsSuffix string, istioClient *crd.Client, k8sClient kubernetes.Interface) *Controller {
+	queue := workqueue.NewRateLimitingQueue(workqueue.DefaultControllerRateLimiter())
```
prabushyam: does ExponentialBackoff rate limiter seem better compared to the default?

pkg/controller/controller.go
```diff
@@ -23,6 +25,11 @@ import (
 	"github.com/yahoo/k8s-athenz-istio-auth/pkg/zms"
 )
 
+const (
+	queueNumRetries = 3
+	queueKey        = "sync"
```
prabushyam: is this queueKey temporary?

pkg/controller/controller.go
```diff
@@ -250,12 +260,45 @@ func (c *Controller) Run(stop chan struct{}) {
...
+func (c *Controller) processNextItem() bool {
+	key, quit := c.queue.Get()
+	if quit {
+		return false
+	}
+
+	defer c.queue.Done(key)
+
+	// TODO, change to new sync function
+	err := c.sync()
```
prabushyam: we might have to pass the key to the `sync()`

pkg/controller/controller.go
mcieplak: @prabushyam This is not needed, will get automatically converted to string.

pkg/controller/controller.go
mcieplak: Internally this is actually calling the exponential backoff starting at 5ms and ending at a 1s max. 
Each retry will be retried after 5(ms)*10^<num-failures>.

pkg/controller/controller.go
mcieplak: Yes, this will be replaced with the athenz domain key once https://github.com/yahoo/k8s-athenz-istio-auth/pull/14 is merged.

pkg/controller/controller.go
mcieplak: Yes, this will go into the logic in a subsequent PR.

PR #19, opened 2019-06-17
Refactor main sync with athenz domain cr as key for work queue

mcieplak:
@prabushyam 

This PR has the following changes:
- Add Athenz Domain CR watch
- Rename store to configStoreCache
- Remove SR / SRB old code, namespace watch, and old sync loop
- AD / SR / SRB events will add a key to the main queue
- Remove zms api code
- Remove poll interval, cert file, key file, and zms url parameters


pkg/controller/controller.go
```diff
@@ -4,296 +4,157 @@
...
-
-			if updated {
-				log.Println("Updated service role binding", roleName, "in namespace", namespace)
-			} else {
-				log.Println("No difference found for service role binding", roleName, "in namespace", namespace,
-					"not updating")
-			}
-		}
+	if !exists {
+		return fmt.Errorf("athenz domain %s does not exist in cache", key)
```
prabushyam: is this an error condition? we might have to delete the CRs for the Athenz domain if they were created at some point.

prabushyam: we could add a TODO, to later add delete enqueue calls to the processing controller

mcieplak: Added a todo comment here so we can handle it when integrating with the processing controller.

PR #20, opened 2019-06-19
Fix the order of model.Config output list and add provider unit tests

prabushyam:
@mcieplak

### DIFF
--- a/pkg/athenz/model_test.go
+++ b/pkg/athenz/model_test.go
@@ -18,10 +18,6 @@ func toRDLTimestamp(s string) (rdl.Timestamp, error) {
 
 func TestGetRolesForDomain(t *testing.T) {
 
-	modified, err := toRDLTimestamp("2018-03-14T19:36:41.003Z")
-	if err != nil {
-		t.Errorf(err.Error())
-	}
 	cases := []struct {
 		test     string
 		domain   *zms.DomainData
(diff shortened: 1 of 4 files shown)


pkg/athenz/model_test.go
```diff
@@ -16,6 +16,75 @@ func toRDLTimestamp(s string) (rdl.Timestamp, error) {
...
+							},
+							{
+								MemberName: "client2-domain.serviceA",
+							},
+						},
+					},
+					{
+						Name:     "athenz-domain.name:role.my-writer-role",
+						Modified: &modified,
+						RoleMembers: []*zms.RoleMember{
```
mcieplak: Should we get rid of the Modified and Role members structs here as they're not actually being used in the unit test?

pkg/athenz/model_test.go
```diff
@@ -431,3 +500,241 @@ func TestGetMembersForRole(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		},
+	}
+
+	for _, c := range cases {
+		if got := ConvertAthenzPoliciesIntoRbacModel(c.domain); !reflect.DeepEqual(got, c.expected) {
+			//assert.True(t, reflect.DeepEqual(c.expected, got), c.test)
```
mcieplak: Can we remove this comment?

pkg/istio/rbac/v1/provider_test.go
```diff
@@ -223,6 +274,197 @@ func TestConvertAthenzModelIntoIstioRbac(t *testing.T) {
...
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+		{
+			test: "model with invalid members that results in unable to create ServiceRoleBindingSpec",
```
mcieplak: Can we say here the test has no members instead of invalid?

PR #21, opened 2019-06-19
Add resync cron for athenz domain and cluster rbac config

mcieplak:
@prabushyam

### DIFF
--- a/main.go
+++ b/main.go
@@ -10,6 +10,7 @@ import (
 	"os/signal"
 	"path/filepath"
 	"syscall"
+	"time"
 
 	"istio.io/istio/pilot/pkg/config/kube/crd"
 	"istio.io/istio/pilot/pkg/model"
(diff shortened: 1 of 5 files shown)


main.go
```diff
@@ -61,18 +64,30 @@ func main() {
...
+	signal.Notify(signalCh, syscall.SIGINT, syscall.SIGTERM)
 	for {
 		select {
-		case <-signalChan:
-			log.Println("Shutdown signal received, exiting...")
-			close(stopChan)
+		case <-signalCh:
+			log.Println("Shutdown signal received, stopping controllers...")
+			close(stopCh)
+			<-signalCh
```
prabushyam: should we simply do a time.Sleep(1s)  instead of waiting for one more SIG?

mcieplak: @prabushyam Sure, I have updated it. The sleep is another convention I noticed people doing here as well.

prabushyam: I am not a fan of time.Sleep either. But, the process should get terminated on SIGTERM, and not wait for SIGKILL IMHO.

PR #23, opened 2019-06-25
Adding processing controller workqueue

prabushyam:
@mcieplak 

- [x] Need to add unit tests for `computeChangeList()`


pkg/controller/controller.go
```diff
@@ -35,13 +38,85 @@ const queueNumRetries = 3
...
+
+	// loop through the desired slice of model.Config and add the items that need to be created or updated
+	for _, desiredConfig := range desired {
+		key := desiredConfig.Key()
+		existingConfig, exists := currMap[key]
+		if !exists {
+			item := processor.Item{
+				Operation:    processor.CREATE,
+				Resource:     desiredConfig,
+				ErrorHandler: nil,
```
mcieplak: Should this be nil here? Don't we need to re-add to the queue?

pkg/controller/controller.go
```diff
@@ -35,13 +38,85 @@ const queueNumRetries = 3
...
+			continue
+		}
+
+		if !equal(existingConfig, desiredConfig) {
+			// copy metadata(for resource version) from current config to desired config
+			desiredConfig.ConfigMeta = existingConfig.ConfigMeta
+			item := processor.Item{
+				Operation:    processor.UPDATE,
+				Resource:     desiredConfig,
+				ErrorHandler: nil,
```
mcieplak: Should this be nil here? Don't we need to re-add to the queue?

pkg/controller/controller.go
```diff
@@ -35,13 +38,85 @@ const queueNumRetries = 3
...
+func equal(c1, c2 model.Config) bool {
+	return c1.Key() == c2.Key() && proto.Equal(c1.Spec, c2.Spec)
+}
+
+// computeChangeList determines a list of change operations to convert the current state of model.Config items into the
+// desired state of model.Config items in the following manner:
+// 1. Converts the current and desired slices into a map for quick lookup
+// 2. Loops through the desired slice of items and identifies items that need to be created/updated
+// 3. Loops through the current slice of items and identifies items that need to be deleted
+func (c *Controller) computeChangeList(current []model.Config, desired []model.Config) []*processor.Item {
```
mcieplak: Does this need to be part of the controller object? Maybe if we replace the nil error handlers we will need to have it to utilize some object?

pkg/istio/onboarding/controller.go
```diff
@@ -5,6 +5,7 @@ package onboarding
 
 import (
 	"errors"
+	"github.com/yahoo/k8s-athenz-istio-auth/pkg/istio/processor"
```
mcieplak: Minor: can we move the import to the bottom?

pkg/istio/onboarding/controller_test.go
```diff
@@ -4,6 +4,7 @@
 package onboarding
 
 import (
+	"github.com/yahoo/k8s-athenz-istio-auth/pkg/istio/processor"
```
mcieplak: Minor: can we move the import to the bottom?

pkg/istio/onboarding/controller_test.go
```diff
@@ -328,6 +337,8 @@ func TestSyncService(t *testing.T) {
 
 			err := c.sync()
 			assert.Nil(t, err, "sync error should be nil")
+			// Add a sleep for processing controller to work on the queue
+			time.Sleep(100 * time.Millisecond)
```
mcieplak: Any better way of doing this without the sleep? Maybe looping until the process queue length is empty?

pkg/istio/processor/controller.go
```diff
@@ -0,0 +1,120 @@
...
+// runWorker calls processNextItem to process events of the work queue
+func (c *Controller) runWorker() {
+	for c.processNextItem() {
+	}
+}
+
+// processNextItem takes an item off the queue and calls the controllers sync
+// function, handles the logic of re-queuing in case any errors occur
+func (c *Controller) processNextItem() bool {
+	itemObj, quit := c.queue.Get()
```
mcieplak: Minor: Can we change itemObj to itemRaw?

pkg/istio/processor/controller_test.go
```diff
@@ -0,0 +1,219 @@
...
+func newCache() (model.ConfigStoreCache, error) {
+	configDescriptor := model.ConfigDescriptor{
+		model.ClusterRbacConfig,
+		model.ServiceRole,
+		model.ServiceRoleBinding,
+	}
+
+	c := memory.NewController(memory.Make(configDescriptor))
+	_, err := c.Create(func() model.Config {
+		return common.NewConfig(model.ServiceRole.Type, "test-ns", "test-role", newSrSpec())
```
mcieplak: Do you think this would be more readable if we changed these one liner function calls to be a temporary variable which is the return from common.NewConfig which is then passed into the Create function? Maybe there's also a way to create two common objects at the top which can be reused in multiple places for service role and service role binding?

pkg/istio/processor/controller_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+	updateTestCache, err := newCache()
+	assert.Nil(t, err, "error should be nil while setting up cache")
+
+	tests := []struct {
+		name          string
+		input         *Item
+		startingCache model.ConfigStoreCache
+		expectedCache model.ConfigStoreCache
+		expectedErr   error
```
mcieplak: Should we add test cases which return the error?

pkg/controller/controller.go
```diff
@@ -144,7 +144,17 @@ func (c *Controller) sync(key string) error {
 	desiredCRs := c.rbacProvider.ConvertAthenzModelIntoIstioRbac(domainRBAC)
 	currentCRs := c.rbacProvider.GetCurrentIstioRbac(domainRBAC, c.configStoreCache)
 
-	changeList := c.computeChangeList(currentCRs, desiredCRs)
+	errHandler := func(err error, item *processor.Item) error {
```
mcieplak: Should we move this outside in order to avoid allocating a new variable every sync call?

pkg/controller/controller_test.go
```diff
@@ -1,6 +1,10 @@
 package controller
 
 import (
+	"fmt"
```
mcieplak: Minor: can we put golang native imports at the top?

pkg/istio/processor/controller_test.go
mcieplak: Could we add this if possible?

pkg/controller/controller.go
```diff
@@ -66,10 +66,10 @@ func equal(c1, c2 model.Config) bool {
 // 1. Converts the current and desired slices into a map for quick lookup
 // 2. Loops through the desired slice of items and identifies items that need to be created/updated
 // 3. Loops through the current slice of items and identifies items that need to be deleted
-func (c *Controller) computeChangeList(current []model.Config, desired []model.Config) []*processor.Item {
+func computeChangeList(current []model.Config, desired []model.Config, errHandler processor.OnErrorFunc) []*processor.Item {
```
mcieplak: I am thinking we make the on error function part of controller and call c.errHandler directly instead of passing it.

PR #24, opened 2019-07-05
Add logging framework, move utils to athenz pkg

mcieplak:
@prabushyam

### DIFF
--- a/main.go
+++ b/main.go
@@ -31,7 +31,7 @@ func main() {
 	adResyncIntervalRaw := flag.String("ad-resync-interval", "1h", "athenz domain resync interval")
 	crcResyncIntervalRaw := flag.String("crc-resync-interval", "1h", "cluster rbac config resync interval")
 	logFile := flag.String("log-file", "/var/log/k8s-athenz-istio-auth/k8s-athenz-istio-auth.log", "log file location")
-	logLevel := flag.String("log-level", "debug", "logging level")
+	logLevel := flag.String("log-level", "info", "logging level")
 
 	flag.Parse()
 	log.InitLogger(*logFile, *logLevel)
--- a/main.go
+++ b/main.go
@@ -90,11 +90,11 @@ func main() {
 	for {
 		select {
 		case <-signalCh:
-			log.Printf("%s Shutdown signal received, stopping controllers...", logPrefix)
+			log.Infof("%s Shutdown signal received, stopping controllers...", logPrefix)
 			close(stopCh)
 			// sleep to allow go routines to successfully exit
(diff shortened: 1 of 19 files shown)


pkg/log/log.go
```diff
@@ -0,0 +1,148 @@
...
+	"path/filepath"
+
+	"github.com/sirupsen/logrus"
+	"gopkg.in/natefinch/lumberjack.v2"
+)
+
+var log *logrus.Logger
+
+// InitLogger initializes a logger object with log rotation
+func InitLogger(logFile, level string) {
```
prabushyam: can we have an empty string as a possible stdout only logging mode? would be useful for tests without having the need to initialize to `/tmp/k8s-athenz-istio-auth.log`

main.go
```diff
@@ -84,11 +90,11 @@ func main() {
 	for {
 		select {
 		case <-signalCh:
-			log.Println("Shutdown signal received, stopping controllers...")
+			log.Printf("%s Shutdown signal received, stopping controllers...", logPrefix)
```
prabushyam: should we update all Printf to corresponding levels like Infof / Warnf / Debugf ?

main.go
```diff
@@ -21,14 +20,21 @@ import (
...
 
+const logPrefix = "[main]"
+
 func main() {
 	dnsSuffix := flag.String("dns-suffix", "svc.cluster.local", "dns suffix used for service role target services")
 	kubeconfig := flag.String("kubeconfig", "", "(optional) absolute path to the kubeconfig file")
 	adResyncIntervalRaw := flag.String("ad-resync-interval", "1h", "athenz domain resync interval")
 	crcResyncIntervalRaw := flag.String("crc-resync-interval", "1h", "cluster rbac config resync interval")
+	logFile := flag.String("log-file", "/var/log/k8s-athenz-istio-auth/k8s-athenz-istio-auth.log", "log file location")
+	logLevel := flag.String("log-level", "debug", "logging level")
```
prabushyam: could we default to info level?

main.go
```diff
@@ -46,32 +52,32 @@ func main() {
 
 	istioClient, err := crd.NewClient(*kubeconfig, "", configDescriptor, *dnsSuffix)
 	if err != nil {
-		log.Panicln("Error creating istio crd client:", err.Error())
+		log.Panicf("%s Error creating istio crd client: %s", logPrefix, err.Error())
```
prabushyam: it would be nice to have a `log.GetPrefixedLogger(prefix string)` that would be a wrapped logger with the prefix, but not sure if possible with our log package implementation

PR #28, opened 2019-07-11
Handle underscore in roles and suppress warnings

prabushyam:
@mcieplak

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -187,7 +187,7 @@ func (c *Controller) sync(key string) error {
 	}
 
 	for _, item := range changeList {
-		log.Infof("Adding resource action to processor queue: %s on %s", item.Operation, item.Resource.Key())
+		log.Infof("Adding resource action to processor queue: %s on %s for key: %s", item.Operation, item.Resource.Key(), key)
 		c.processor.ProcessConfigChange(item)
 	}
 
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -15,6 +15,7 @@ import (
 	"istio.io/istio/pilot/pkg/serviceregistry/kube"
 
 	"k8s.io/api/core/v1"
+	apiErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
@@ -127,6 +128,14 @@ func (c *Controller) getCallbackHandler(key string) processor.OnCompleteFunc {
 		}
 		if item != nil {
 			log.Errorf("Error performing %s on %s: %s", item.Operation, item.Resource.Key(), err.Error())
+		}
+		if apiErrors.IsNotFound(err) || apiErrors.IsAlreadyExists(err) {
+			log.Infof("Error is non-retryable %s", err)
+			return nil
+		}
+		if !apiErrors.IsConflict(err) {
+			log.Infof("Retrying operation %s on %s due to processing error for %s", item.Operation, item.Resource.Key(), key)
+			return err
 		}
 		if c.queue.NumRequeues(key) >= queueNumRetries {
 			log.Errorf("Max number of retries reached for %s.", key)
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -67,7 +67,7 @@ func equal(c1, c2 model.Config) bool {
 // 1. Converts the current and desired slices into a map for quick lookup
 // 2. Loops through the desired slice of items and identifies items that need to be created/updated
 // 3. Loops through the current slice of items and identifies items that need to be deleted
-func computeChangeList(current []model.Config, desired []model.Config, errHandler processor.OnCompleteFunc) []*processor.Item {
+func computeChangeList(current []model.Config, desired []model.Config, cbHandler processor.OnCompleteFunc) []*processor.Item {
 
 	currMap := convertSliceToKeyedMap(current)
 	desiredMap := convertSliceToKeyedMap(desired)
@@ -82,7 +82,7 @@ func computeChangeList(current []model.Config, desired []model.Config, errHandle
 			item := processor.Item{
 				Operation:       model.EventAdd,
 				Resource:        desiredConfig,
-				CallbackHandler: errHandler,
+				CallbackHandler: cbHandler,
 			}
 			changeList = append(changeList, &item)
 			continue
@@ -94,7 +94,7 @@ func computeChangeList(current []model.Config, desired []model.Config, errHandle
 			item := processor.Item{
 				Operation:       model.EventUpdate,
 				Resource:        desiredConfig,
-				CallbackHandler: errHandler,
+				CallbackHandler: cbHandler,
 			}
 			changeList = append(changeList, &item)
 			continue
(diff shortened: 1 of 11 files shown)


pkg/controller/controller_test.go
```diff
@@ -241,7 +241,7 @@ func TestComputeChangeList(t *testing.T) {
 	type input struct {
 		current    []model.Config
 		desired    []model.Config
-		errHandler processor.OnErrorFunc
+		errHandler processor.OnCompleteFunc
```
mcieplak: Can we also update the errHandler variable name?

pkg/istio/onboarding/controller.go
```diff
@@ -98,7 +98,6 @@ func (c *Controller) processNextItem() bool {
 		}
 	}
 
-	c.queue.Forget(key)
```
mcieplak: We no longer have a Forget call anywhere in the onboarding controller now. I believe we need to have similar logic as the main controller where we forget after a sync has a no-op.

pkg/istio/onboarding/controller.go
```diff
@@ -181,14 +180,22 @@ func (c *Controller) getOnboardedServiceList() []string {
...
+// callbackHandler re-adds the key for a failed processor.sync operation
+func (c *Controller) callbackHandler(err error, item *processor.Item) error {
+	if err == nil {
+		return nil
+	}
+	if item != nil {
+		log.Errorf("%s Error performing %s on %s: %s", logPrefix, item.Operation, item.Resource.Key(), err)
+	}
+	if c.queue.NumRequeues(queueKey) >= queueNumRetries {
+		log.Errorf("%s Max number of retries reached for %s.", logPrefix, queueKey)
```
mcieplak: I think we need alerts for this case since there is only one key when it comes to this controller.

pkg/istio/processor/controller.go
```diff
@@ -22,12 +22,15 @@ type Controller struct {
...
 
 type Item struct {
-	Operation    model.Event
-	Resource     model.Config
-	ErrorHandler OnErrorFunc
+	Operation model.Event
+	Resource  model.Config
+
+	// Handler function that should be invoked with the status of the current sync operation on the item
+	// If the handler returns an error, the operation is retried upto `queueNumRetries`
```
mcieplak: Minor typo: `upto` -> `up to`

pkg/istio/processor/controller.go
```diff
@@ -80,16 +83,23 @@ func (c *Controller) processNextItem() bool {
...
-			}
-		}
+	}
+	if item.CallbackHandler == nil {
+		c.queue.Forget(itemRaw)
+		return true
 	}
 
+	// All errors/successes should be handled by the CallbackHandler()
+	err = item.CallbackHandler(err, item)
```
mcieplak: Currently both of the controllers never return an error here. As we discussed we should add explicit error checks and retry here on network issues and potentially some api server errors.

pkg/istio/processor/controller_test.go
```diff
@@ -110,9 +110,9 @@ func TestSync(t *testing.T) {
 		{
 			name: "should perfom valid create operation",
 			input: &Item{
-				Operation:    model.EventAdd,
-				Resource:     newSr("test-ns", "test-role"),
-				ErrorHandler: errHandler,
+				Operation:       model.EventAdd,
+				Resource:        newSr("test-ns", "test-role"),
+				CallbackHandler: errHandler,
```
mcieplak: Can we also update the errHandler variable here?

pkg/istio/rbac/common/servicerolebinding.go
```diff
@@ -37,7 +37,7 @@ func parseMemberName(member *zms.RoleMember) (string, error) {
 }
 
 // GetServiceRoleBindingSpec returns the ServiceRoleBindingSpec for a given Athenz role and its members
-func GetServiceRoleBindingSpec(roleName string, members []*zms.RoleMember) (*v1alpha1.ServiceRoleBinding, error) {
+func GetServiceRoleBindingSpec(roleName string, k8sRoleName string, members []*zms.RoleMember) (*v1alpha1.ServiceRoleBinding, error) {
```
mcieplak: Do we still need the roleName variable?

pkg/controller/controller.go
```diff
@@ -130,12 +131,20 @@ func (c *Controller) getCallbackHandler(key string) processor.OnCompleteFunc {
 		if item != nil {
 			log.Errorf("%s Error performing %s on %s: %s", logPrefix, item.Operation, item.Resource.Key(), err)
 		}
+		if apiErrors.IsNotFound(err) || apiErrors.IsAlreadyExists(err) {
```
mcieplak: Why are these not retryable?

pkg/controller/controller.go
```diff
@@ -117,16 +118,33 @@ func computeChangeList(current []model.Config, desired []model.Config, errHandle
...
+
+		if err == nil {
+			return nil
+		}
+		if item != nil {
+			log.Errorf("Error performing %s on %s: %s", item.Operation, item.Resource.Key(), err.Error())
+		}
+		if apiErrors.IsNotFound(err) || apiErrors.IsAlreadyExists(err) {
+			log.Infof("Error is non-retryable %s", err)
+			return nil
```
mcieplak: Should we not add to the queue here?

pkg/controller/controller.go
```diff
@@ -106,9 +107,9 @@ func computeChangeList(current []model.Config, desired []model.Config, errHandle
 		_, exists := desiredMap[key]
 		if !exists {
 			item := processor.Item{
-				Operation:    model.EventDelete,
-				Resource:     currConfig,
-				ErrorHandler: errHandler,
+				Operation:       model.EventDelete,
+				Resource:        currConfig,
+				CallbackHandler: errHandler,
```
mcieplak: Can we rename all these `errHandler` to `cbHandler` or something along those lines everywhere?

pkg/controller/controller.go
prabushyam: This is a case, where say a CREATE got scheduled, but another CREATE created the resource and API returned AlreadyExists. Now, the change for the CREATE would trigger another sync that should take care of further changes if any needed, so we wouldn't have to add it back to the queue in this case.

pkg/istio/onboarding/controller.go
```diff
@@ -180,14 +180,30 @@ func (c *Controller) getOnboardedServiceList() []string {
...
+func (c *Controller) callbackHandler(err error, item *processor.Item) error {
+	if err == nil {
+		return nil
+	}
+	if item != nil {
+		log.Errorf("Error performing %s on %s: %s", item.Operation, item.Resource.Key(), err)
+	}
+	if apiErrors.IsNotFound(err) || apiErrors.IsAlreadyExists(err) {
+		log.Infof("Error is non-retryable %s", err)
+		return nil
```
mcieplak: Shouldn't this trigger a new resync of the controller?

pkg/controller/controller.go
prabushyam: oops, my bad. I thought I had updated it.

pkg/istio/onboarding/controller.go
prabushyam: Similar logic goes here.. since the change already went through by another operation, and that would anyway trigger another sync, we do not have to retry/re-add this error

pkg/controller/controller.go
mcieplak: Ah got it, that sounds good.

pkg/istio/onboarding/controller.go
mcieplak: That makes sense, thanks.

PR #36, opened 2019-08-24
Add service role / binding integration tests

mcieplak:
@prabushyam 

Still finishing up / cleaning up a few small things in this PR but it is mainly ready for review. Please take an initial look at the approach and we can go from there.


test/integration/authz_v1_test.go
```diff
@@ -0,0 +1,388 @@
...
+
+			if !reflect.DeepEqual(curr.Spec, got.Spec) {
+				return false, nil
+			}
+		}
+
+		return true, nil
+	})
+
+	if err != nil {
```
prabushyam: we could follow `assert.Nil(...` pattern as in other places

test/integration/authz_v1_test.go
```diff
@@ -0,0 +1,388 @@
...
+		t.Error("time out waiting for rollout for ad", r.AD.Name, "with error", err)
+	}
+
+	validateConfigs(t, r)
+}
+
+// validateConfigs will validate the service role / service role bindings on the
+// cluster against the expected output
+func validateConfigs(t *testing.T, r *fixtures.ExpectedResources) {
+	for _, curr := range r.ModelConfigs {
```
prabushyam: could we also validate if there are no other resources expect the ones expected in `r.ModelConfigs`?

test/integration/authz_v1_test.go
```diff
@@ -0,0 +1,388 @@
...
+		assert.NotNil(t, got, "istio custom resource should exist on the cluster")
+		curr.ResourceVersion = got.ResourceVersion
+		curr.CreationTimestamp = got.CreationTimestamp
+		assert.Equal(t, curr, *got, "istio resource should be equal to expected")
+	}
+}
+
+// cleanup will clean up the athenz domain and service role / service role binding objects on the cluster
+func cleanup(t *testing.T, r *fixtures.ExpectedResources) {
+	err := framework.Global.AthenzDomainClientset.AthenzV1().AthenzDomains().Delete(r.AD.Name, &v1.DeleteOptions{})
```
prabushyam: should we do a List() and delete all AD, SR, SRB, CRC resources? (if controller created unexpected resources, they might not get cleaned up

test/integration/authz_v1_test.go
```diff
@@ -0,0 +1,388 @@
...
+func TestCreateServiceRoleOnly(t *testing.T) {
+	o := &fixtures.OverrideResources{
+		ModifyAD: func(signedDomain *zms.SignedDomain) {
+			signedDomain.Domain.Roles[0].RoleMembers = []*zms.RoleMember{}
+			signedDomain.Domain.Roles[0].Members = []zms.MemberName{}
+		},
+	}
+	r := fixtures.GetExpectedResources(o)
+	rolloutAndValidate(t, r, create)
+	srb := framework.Global.IstioClientset.Get(model.ServiceRoleBinding.Type, "client-writer-role", r.AD.Namespace)
```
prabushyam: should `r.AD.Namespace` be `athenz.DomainToNamespace(r.AD.Name)` ?

test/integration/authz_v1_test.go
```diff
@@ -0,0 +1,388 @@
...
+	assert.Nil(t, srb, "service role binding should not exist")
+	cleanup(t, r)
+}
+
+// 2.0 Update existing role / policy
```
prabushyam: should this be `Update existing domain with new role / policy`?

test/integration/authz_v1_test.go
```diff
@@ -0,0 +1,388 @@
...
+			signedDomain.Domain.Policies.Contents.Policies[0].Assertions[0].Role = "athenz.domain.two:role.client-reader-role"
+			signedDomain.Domain.Policies.Contents.Policies[0].Assertions[0].Resource = "athenz.domain.two:svc.my-service-name"
+			signedDomain.Domain.Roles[0].Name = "athenz.domain.two:role.client-reader-role"
+		},
+	}
+
+	rThree := fixtures.GetExpectedResources(o)
+	rolloutAndValidate(t, rThree, create)
+
+	cleanup(t, rOne)
```
prabushyam: could we add one update action to a role/policy on one of the AD as the test case indicates?

test/integration/authz_v1_test.go
```diff
@@ -0,0 +1,388 @@
...
+
+// 3.0 Delete athenz domain
+// TODO: currently the controller does not delete the service role / binding
+// due to checking for the existence of the athenz domain. This test assumes
+// these still exist. Update this test once the controller core logic changes.
+func TestAthenzDomainDelete(t *testing.T) {
+	r := fixtures.GetExpectedResources(&fixtures.OverrideResources{})
+	rolloutAndValidate(t, r, create)
+
+	err := framework.Global.AthenzDomainClientset.AthenzV1().AthenzDomains().Delete(r.AD.Name, &v1.DeleteOptions{})
```
prabushyam: would be nice if we can make `rolloutAndValidate` handle `delete` similar to `create` & `update` as well

test/integration/framework/framework.go
```diff
@@ -118,13 +128,13 @@ func Setup() error {
 		model.ClusterRbacConfig,
 	}
 
-	istioClientset, err := crd.NewClient("", "", configDescriptor, "svc.cluster.local")
```
prabushyam: why remove?

test/integration/fixtures/fixtures.go
```diff
@@ -165,57 +172,93 @@ func CreateCrds(clientset *apiextensionsclient.Clientset) error {
...
+	ModifyAD           func(signedDomain *zms.SignedDomain)
+	ModifySRAndSRBPair []func(sr *v1alpha1.ServiceRole, srb *v1alpha1.ServiceRoleBinding)
+}
+
+// GetExpectedResources returns an expected resources object which contains the
+// athenz domain along with its service roles / bindings objects
+func GetExpectedResources(o *OverrideResources) *ExpectedResources {
+	signedDomain := getDefaultSignedDomain()
+
+	if o.ModifyAD != nil {
```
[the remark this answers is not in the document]
mcieplak: Yep, added the check to be safe.

test/integration/fixtures/fixtures.go
```diff
@@ -228,14 +271,13 @@ func getFakeDomain() zms.SignedDomain {
 						{
 							Assertions: []*zms.Assertion{
 								{
-									Role:     domainName + ":role.admin",
-									Resource: domainName + ".test:*",
-									Action:   "*",
 									Effect:   &allow,
+									Action:   "put",
+									Role:     "athenz.domain:role.client-writer-role",
```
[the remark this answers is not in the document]
mcieplak: Sure, updated.

test/integration/authz_v1_test.go
mcieplak: Good catch, updated.

test/integration/authz_v1_test.go
mcieplak: Yep, I added a check.

test/integration/authz_v1_test.go
mcieplak: Yes, I added a check that the list must return empty for that domain / namespace after the deletes go through. This will catch any extra resources.

test/integration/authz_v1_test.go
mcieplak: Ah yes, this is fixed now.

test/integration/authz_v1_test.go
mcieplak: Fixed.

test/integration/authz_v1_test.go
mcieplak: I added an update to all of them.

test/integration/authz_v1_test.go
mcieplak: That's a great point, updated it.

test/integration/framework/framework.go
mcieplak: Added it back now.

test/integration/authz_v1_test.go
mcieplak: Yes, added it.

mcieplak:
@prabushyam Updated based on your feedback, please review.

test/integration/authz_v1_test.go
mcieplak: The CRC will be handled in the subsequent PR.

PR #42, opened 2020-01-15
Add support for jwt authz in srb

mcieplak:
@prabushyam @sureshvis

### DIFF
--- a/pkg/istio/rbac/common/servicerolebinding.go
+++ b/pkg/istio/rbac/common/servicerolebinding.go
@@ -20,59 +20,82 @@ const (
 	RequestAuthPrincipalProperty = "request.auth.principal"
 )
 
-// parseMemberName parses the Athenz role member into a SPIFFE compliant name
-func parseMemberName(member *zms.RoleMember) (string, string, error) {
+// memberToSpiffe parses the Athenz role member into a SPIFFE compliant name.
+// Example: example.domain/sa/service
+func memberToSpiffe(member *zms.RoleMember) (string, error) {
 
 	if member == nil {
-		return "", "", fmt.Errorf("member is nil")
+		return "", fmt.Errorf("member is nil")
 	}
 
 	memberStr := string(member.MemberName)
 
 	// special condition: if member == 'user.*', return '*'
 	if memberStr == allUsers {
-		return WildCardAll, WildCardAll, nil
+		return WildCardAll, nil
 	}
 
-	spiffeName, err := PrincipalToSpiffe(memberStr)
-	if err != nil {
-		return "", "", err
+	return PrincipalToSpiffe(memberStr)
+}
+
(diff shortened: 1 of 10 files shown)


pkg/istio/rbac/common/servicerolebinding.go
```diff
@@ -13,26 +13,34 @@ import (
...
-	WildCardAll     = "*"
-	ServiceRoleKind = "ServiceRole"
+	allUsers                     = "user.*"
+	WildCardAll                  = "*"
+	ServiceRoleKind              = "ServiceRole"
+	AthenzJwtPrefix              = "athenz/"
+	RequestAuthPrincipalProperty = "request.auth.principal"
 )
 
 // parseMemberName parses the Athenz role member into a SPIFFE compliant name
```
prabushyam: we could update the function comment to include the JWT format as well

pkg/istio/rbac/common/servicerolebinding.go
```diff
@@ -43,17 +51,28 @@ func GetServiceRoleBindingSpec(k8sRoleName string, members []*zms.RoleMember) (*
...
+		// Spiffe and request auth principal subjects MUST be separate or else
+		// the user needs to provide both the certificate and the jwt token. If
+		// one subject is used, the source.principal and request.auth.principal
+		// in the envoy rbac is grouped together into one id principal array as
+		// opposed to being separated to allow either to go through.
+		spiffeSubject := &v1alpha1.Subject{
+			User: spiffeName,
+		}
+
+		requestAuthPrincipalSubject := &v1alpha1.Subject{
```
prabushyam: could we have the jwt subject added based on a cmd line flag (e.g. `--enable-origin-jwt-subject` ) or a better name, so that we can turn on/off quickly. In this case, `parseMemberName` could be separated and renamed to something like `memberToSpiffe` and the jwt one to`memberToOriginSubject`. What do you think?

mcieplak: I agree we should add the feature flag in case we want to roll back. It's a little bit tricky because this function does not have access to much besides its parameters. I am thinking of two ways of doing this:
1. Add the flag to the provider object which will pass it as a parameter to GetServiceRoleBindingSpec function
2.  Introduce the feature flag global object used in kubernetes. This might be overkill as we only have one flag for now but we could potentially have more in the future.

sureshvis: +1 to feature flag it

sureshvis: creating a global object in k8s seems overkill for feature flag.

mcieplak: @prabushyam @prabushyam I have added the flag and also separated the functions. Please review.

PR #47, opened 2021-01-06
Authz Policy Controller

RuiGuo34:
@mcieplak @mdhume please review. 


pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,549 @@
...
+				// case when service has authzEnabledAnnotation removed, and authz policy with the same name present
+				if res := c.configStoreCache.Get(schemas.AuthorizationPolicy.Type, obj.Name, obj.Namespace); res != nil {
+					log.Infoln("Deleting Authz Policy ... ")
+					c.configStoreCache.Delete(schemas.AuthorizationPolicy.Type, obj.Name, obj.Namespace)
+				}
+			}
+		} else if castItem.Operation == model.EventUpdate {
+			if _, ok = obj.Annotations[authzEnabledAnnotation]; ok {
+				if obj.Annotations[authzEnabledAnnotation] == authzEnabled {
+					if res := c.configStoreCache.Get(schemas.AuthorizationPolicy.Type, obj.Name, obj.Namespace); res == nil {
```
mcieplak, line 224: Can we combine this logic with the create operation? We could make this a common function.

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,549 @@
...
+		} else if castItem.Operation == model.EventDelete {
+			log.Infoln("Deleting Authz Policy ... ")
+			err = c.configStoreCache.Delete(schemas.AuthorizationPolicy.Type, obj.Name, obj.Namespace)
+			if err != nil {
+				return err
+			}
+		}
+	} else if obj, ok := (castItem.Resource).(*adv1.AthenzDomain); ok {
+		// athenz domain update will result in update in the authz policies in the corresponding namespace
+		if castItem.Operation == model.EventUpdate {
```
mcieplak, line 278: Aren't we missing the create / delete events here as well?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,549 @@
...
+			}
+		}
+	} else {
+		return fmt.Errorf("unable to cast interface to service or athenzDomain object")
+	}
+	return nil
+}
+
+// checkAuthzEnabledAnnotation checks if current servce object has "authz.istio.io/enabled" annotation set
+func (c *Controller) checkAuthzEnabledAnnotation(serviceObj *corev1.Service) bool {
```
mcieplak, line 310: It looks like this function is not being used?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,549 @@
...
+	if _, ok := serviceObj.Annotations[authzEnabledAnnotation]; ok {
+		return true
+	}
+	return false
+}
+
+func (c *Controller) convertAthenzModelIntoIstioAuthzPolicy(athenzModel athenz.Model, namespace string, serviceName string, svcLabel string) model.Config {
```
mcieplak, line 317: Can we move this logic into a v2 folder under https://github.com/yahoo/k8s-athenz-istio-auth/tree/master/pkg/istio/rbac which implements the provider interface here https://github.com/yahoo/k8s-athenz-istio-auth/blob/master/pkg/istio/rbac/interface.go?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,549 @@
...
+			// assert.Resource contains the svc information that needs to parse and match
+			svc, path, err := parseAssertionResource(zms.DomainName(m.NamespaceToDomain(namespace)), assert)
+			if err != nil {
+				continue
+			}
+			// if svc match with current svc, process it and add it to the rules
+			// note that svc defined on athenz can be a regex, need to match the pattern
+			res, e := regexp.MatchString(svc, svcLabel)
+			if e != nil {
+				log.Errorln("error matching string: ", e.Error())
```
mcieplak, line 352: Should we add a continue here?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,549 @@
...
+						// add function to enableOriginJwtSubject, following code assume enableOriginJwtSubject is true by default
+						if c.enableOriginJwtSubject {
+							for _, roleMember := range athenzModel.Members[roleName] {
+								spiffeName, err := memberToSpiffe(roleMember)
+								if err != nil {
+									log.Errorln("error converting role name to spiffeName: ", err.Error())
+									continue
+								}
+								from_principal.Source.Principals = append(from_principal.Source.Principals, spiffeName)
+								originJwtName, err := memberToOriginJwtSubject(roleMember)
```
mcieplak, line 376: Should this logic be around the `if c.enableOriginJwtSubject {` statement? We want to add the source principals above regardless of the enableOriginJwtSubject flag.

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,549 @@
...
+		}
+	}
+	spec.Rules = rules
+	out.Spec = spec
+	return out
+}
+
+// parseAssertionResource parses the resource of an action into the service name (AccessRule constraint) and the
+// HTTP paths if specified (suffix :<path>)
+func parseAssertionResource(domainName zms.DomainName, assertion *zms.Assertion) (string, string, error) {
```
mcieplak, line 429: Can we reuse all the functions in the common package instead of having them here? We can move them to different files if they're currently in the service role / binding files.

mcieplak (changes requested):
@RuiGuo34 It looks like the travis ci build is also failing

go.mod
```diff
@@ -4,18 +4,61 @@ go 1.12
...
+	istio.io/istio v0.0.0-20200220235129-f288658b710d
+	istio.io/pkg v0.0.0-20191030005435-10d06b6b315e
+	k8s.io/api v0.18.0
+	k8s.io/apimachinery v0.18.0
+	k8s.io/client-go v15.0.0+incompatible
 )
+
+// Kubernetes makes it challenging to depend on their libraries. To get around this, we need to force
+// the sha to use. All of these are pinned to the tag "kubernetes-1.16"
+replace k8s.io/api => k8s.io/api v0.0.0-20191003000013-35e20aa79eb8
```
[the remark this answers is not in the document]
RuiGuo34: I made an update to remove all the replaces as many as possible.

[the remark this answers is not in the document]
RuiGuo34: Left replace is there because istio 1.5.8 library pulls k8s dependencies higher than 1.18 (go client, api, etc..), pinned to version 17.14 which cluster will be upgraded to.

main.go
```diff
@@ -35,11 +37,12 @@ func main() {
 	flag.Parse()
 	log.InitLogger(*logFile, *logLevel)
 
-	configDescriptor := model.ConfigDescriptor{
-		model.ServiceRole,
-		model.ServiceRoleBinding,
-		model.ClusterRbacConfig,
-	}
+	// configDescriptor := model.ConfigDescriptor{
```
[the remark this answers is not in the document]
RuiGuo34: removed

main.go
```diff
@@ -49,7 +52,9 @@ func main() {
 		}
 	}
 
-	istioClient, err := crd.NewClient(*kubeconfig, "", configDescriptor, *dnsSuffix)
+	// Ledger for tracking config distribution, specify how long it can retain its previous state
+	configLedger := ledger.Make(time.Hour)
```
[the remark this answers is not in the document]
RuiGuo34: this is the package overview: https://pkg.go.dev/istio.io/pkg/ledger. It is a key value store which provides a unique and reproducible version identifier for every combination of keys and values it has seen.

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,549 @@
+// Copyright 2019, Verizon Media Inc.
```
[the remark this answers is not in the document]
RuiGuo34: updated

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,549 @@
...
+		DeleteFunc: func(obj interface{}) {
+			item := Item{
+				Operation: model.EventDelete,
+				Resource:  obj,
+			}
+			c.ProcessConfigChange(item)
+		},
+	})
+
+	adIndexInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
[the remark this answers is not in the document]
RuiGuo34: added delete event

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,549 @@
...
+	"istio.io/istio/pilot/pkg/model"
+	"istio.io/istio/pkg/config/constants"
+	"istio.io/istio/pkg/config/schemas"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+)
+
+const (
```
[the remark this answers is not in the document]
RuiGuo34: moved them to a common file

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,549 @@
...
+	queueNumRetries        = 3
+	authzEnabled           = "true"
+	authzEnabledAnnotation = "authz.istio.io/enabled"
+)
+
+var resourceRegex = regexp.MustCompile(`\A(?P<domain>.*):svc.(?P<svc>[^:]*)[:]?(?P<path>.*)\z`)
+
+var supportedMethods = map[string]bool{
```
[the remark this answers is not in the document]
RuiGuo34: moved to a common file

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,549 @@
...
+
+// runWorker calls processNextItem to process events of the work queue
+func (c *Controller) runWorker() {
+	for c.processNextItem() {
+	}
+}
+
+// processNextItem takes an item off the queue and calls the controllers sync
+// function, handles the logic of re-queuing in case any errors occur
+func (c *Controller) processNextItem() bool {
```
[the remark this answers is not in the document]
RuiGuo34: added logic and changed if block to switch case for better readability

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,549 @@
...
+// sync is responsible for invoking the appropriate API operation on the model.Config resource
+func (c *Controller) sync(item interface{}) error {
+	var err error
+	castItem, ok := item.(Item)
+	if !ok {
+		return fmt.Errorf("Unable to cast interface to service object")
+	}
+	// dealing with service resource
+
+	if obj, ok := (castItem.Resource).(*corev1.Service); ok {
```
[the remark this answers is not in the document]
RuiGuo34: added logic and changed if block to switch case for better readability

pkg/istio/authorizationpolicy/controller.go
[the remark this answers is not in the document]
RuiGuo34: added

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,549 @@
...
+				} else {
+					// case when service has authz flag switch from true to false, authz policy with the same name present
+					if res := c.configStoreCache.Get(schemas.AuthorizationPolicy.Type, obj.Name, obj.Namespace); res != nil {
+						log.Infoln("Deleting Authz Policy ... ")
+						c.configStoreCache.Delete(schemas.AuthorizationPolicy.Type, obj.Name, obj.Namespace)
+					}
+				}
+			} else {
+				// case when service has authzEnabledAnnotation removed, and authz policy with the same name present
+				if res := c.configStoreCache.Get(schemas.AuthorizationPolicy.Type, obj.Name, obj.Namespace); res != nil {
```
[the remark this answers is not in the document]
RuiGuo34: combined logic

pkg/istio/authorizationpolicy/controller.go
RuiGuo34: created a common function for create

pkg/istio/authorizationpolicy/controller.go
RuiGuo34: delete events are rarely seen for athenz domain crd. When there is such event, it usually means namespace is deleted, all the resources with the namespace are deleted. We can think of the use case if we want to have delete events added here later.

pkg/istio/authorizationpolicy/controller.go
RuiGuo34: it is being used to check before authz policy create/update

pkg/istio/authorizationpolicy/controller.go
RuiGuo34: moved.

pkg/istio/authorizationpolicy/controller.go
RuiGuo34: added, thanks

pkg/istio/authorizationpolicy/controller.go
RuiGuo34: fixed the logic

pkg/istio/authorizationpolicy/controller.go
RuiGuo34: moved to a common package

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+				if castItem.Operation == model.EventUpdate {
+					// assign current revision, update function requires a defined resource version
+					convertedCR.ResourceVersion = obj.ResourceVersion
+					_, err := c.ConfigStoreCache.Update(convertedCR)
+					if err != nil {
+						log.Errorln("error updating authz policy: ", err.Error())
+						return err
+					}
+				}
+				// prevent
```
mcieplak, line 362: Can we remove this comment or expand upon it?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+	}
+	return nil
+}
+
+// checkAuthzEnabledAnnotation checks if current servce object has "authz.istio.io/enabled" annotation set
+func (c *Controller) checkAuthzEnabledAnnotation(serviceObj *corev1.Service) bool {
+	if _, ok := serviceObj.Annotations[authzEnabledAnnotation]; ok {
+		if serviceObj.Annotations[authzEnabledAnnotation] == authzEnabled {
+			return true
+		} else {
```
mcieplak, line 386: We could remove this else block as we directly return false.

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+			return false
+		}
+	}
+	return false
+}
+
+func (c *Controller) createDryrunResource(convertedCR model.Config, authzPolicyName string, namespace string) error {
+	convertedObj, err := crd.ConvertConfig(collections.IstioSecurityV1Beta1Authorizationpolicies, convertedCR)
+	if err != nil {
+		log.Errorln("Unable to convert authorization policy config to istio objects")
```
mcieplak, line 396: Shouldn't we return the error here?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+
+func (c *Controller) findDeleteDryrunResource(authzPolicyName string, namespace string) error {
+	yamlFileName := authzPolicyName + "--" + namespace + ".yaml"
+	if _, err := os.Stat(DryRunStoredFilesDirectory + yamlFileName); os.IsNotExist(err) {
+		log.Infof("file %s does not exist in local directory\n", DryRunStoredFilesDirectory + yamlFileName)
+		return nil
+	}
+	log.Infof("deleting file under path: %s\n", DryRunStoredFilesDirectory + yamlFileName)
+	err := os.Remove(DryRunStoredFilesDirectory + yamlFileName)
+
```
mcieplak, line 418: Minor: could we remove this newline?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+}
+
+func (c *Controller) createAuthzPolicyResource(convertedCR model.Config, obj *corev1.Service) error {
+	// form the authorization policy config and send create sign to the queue
+	athenzDomainRaw, exists, err := c.AdIndexInformer.GetIndexer().GetByKey(athenz.NamespaceToDomain(obj.Namespace))
+	if err != nil {
+		return err
+	}
+	if !exists {
+		// TODO, add the non existing athenz domain to the istio custom resource
```
mcieplak, line 432: Should we remove this comment?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+	convertedObj, err := crd.ConvertConfig(collections.IstioSecurityV1Beta1Authorizationpolicies, convertedCR)
+	if err != nil {
+		log.Errorln("Unable to convert authorization policy config to istio objects")
+	}
+	configInBytes, err := yaml.Marshal(convertedObj)
+	if err != nil {
+		return fmt.Errorf("could not marshal %v: %v", convertedCR.Name, err)
+	}
+	yamlFileName := authzPolicyName + "--" + namespace + ".yaml"
+	err = ioutil.WriteFile(DryRunStoredFilesDirectory + yamlFileName, configInBytes, 0644)
```
mcieplak, line 403: Should we directly return with the WriteFile call? We have the log line with the "unable write to file" string everywhere we call this function and it returns an error.

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+	yamlFileName := authzPolicyName + "--" + namespace + ".yaml"
+	if _, err := os.Stat(DryRunStoredFilesDirectory + yamlFileName); os.IsNotExist(err) {
+		log.Infof("file %s does not exist in local directory\n", DryRunStoredFilesDirectory + yamlFileName)
+		return nil
+	}
+	log.Infof("deleting file under path: %s\n", DryRunStoredFilesDirectory + yamlFileName)
+	err := os.Remove(DryRunStoredFilesDirectory + yamlFileName)
+
+	if err != nil {
+		return fmt.Errorf("unable to delete file, err: %v", err)
```
mcieplak, line 420: Should we directly return with the Remove call? We have the log line which mentions that we couldn't delete the file in case of an error wherever this function is called.

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+	// check if resource can be cast to service, athenzdomain, or authorizationpolicy object
+	_, svcCast := (castItem.Resource).(*corev1.Service)
+	_, adCast := (castItem.Resource).(*adv1.AthenzDomain)
+	_, apCast := (castItem.Resource).(*securityv1beta1.AuthorizationPolicy)
+	if !(svcCast || adCast || apCast) {
+		return fmt.Errorf("unable to cast interface to service or athenzDomain or authz policy object")
+	}
+
+	// dealing with service resource
+	switch true {
```
mcieplak, line 204: I'm wondering if we should separate each of these event handlers into separate functions so that this function is just a bunch of function calls. It might make it easier to read. What do you think?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+				if err != nil {
+					log.Errorln("error deleting local authz policy file: ", err.Error())
+					return err
+				}
+			}
+		}
+	case adCast:
+		obj, _ := (castItem.Resource).(*adv1.AthenzDomain)
+		// athenz domain create/update event should trigger a sync with existing authz policies in the corresponding namespace
+		if castItem.Operation == model.EventUpdate || castItem.Operation == model.EventAdd {
```
mcieplak, line 297: Can we also add the delete logic?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+	log.Infof("deleting file under path: %s\n", DryRunStoredFilesDirectory + yamlFileName)
+	err := os.Remove(DryRunStoredFilesDirectory + yamlFileName)
+
+	if err != nil {
+		return fmt.Errorf("unable to delete file, err: %v", err)
+	}
+	return nil
+}
+
+func (c *Controller) createAuthzPolicyResource(convertedCR model.Config, obj *corev1.Service) error {
```
mcieplak, line 425: I see there's some duplication with this function, the athenz domain create / update, and the authz policy events. I'm wondering if there's a way to nicely combine this logic.

pkg/istio/common/common_test.go
```diff
@@ -0,0 +1,47 @@
+package common
```
mcieplak, line 1: Can we move the TestMemberToSpiffe, TestMemberToOriginJwtSubject, TestParseAssertionEffect, TestParseAssertionAction, and TestParseAssertionResource tests to this file as well?

pkg/istio/common/common.go
```diff
@@ -0,0 +1,162 @@
+// Copyright 2021, Verizon Media Inc.
+// Licensed under the terms of the 3-Clause BSD license. See LICENSE file in
+// github.com/yahoo/k8s-athenz-istio-auth for terms.
+package common
```
mcieplak, line 4: Can we combine this file with the util.go file? One file with the common functionality should be enough.

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+				authzSpec, ok := (authzPolicy.Spec).(*v1beta1.AuthorizationPolicy)
+				if !ok {
+					return fmt.Errorf("unable to cast interface to authorizationpolicies object")
+				}
+				convertedCR := c.RbacProviderV2.ConvertAthenzModelIntoIstioAuthzPolicy(domainRBAC, authzPolicy.Namespace, authzPolicy.Name, authzSpec.Selector.MatchLabels["svc"])
+				log.Infof("Athenz Domain %s updated, updating Authz Policy in namespace %s ... ", obj.Name, m.DomainToNamespace(obj.Name))
+				// assign current revision, update function requires a defined resource version
+				convertedCR.ResourceVersion = authzPolicy.ResourceVersion
+				if (!c.Dryrun) {
+					revision, err := c.ConfigStoreCache.Update(convertedCR)
```
mcieplak, line 314: We are updating every single resource here even if it didn't have a change. In the existing controller, we calculate only the items that changed so that we can apply them: https://github.com/yahoo/k8s-athenz-istio-auth/blob/master/pkg/controller/controller.go#L180
Can we do the same here? This will avoid a lot of extra api calls.

pkg/istio/rbac/interface.go
```diff
@@ -18,3 +18,9 @@ type Provider interface {
 	// GetCurrentIstioRbac returns the Istio RBAC custom resources associated with the given model
 	GetCurrentIstioRbac(model athenz.Model, csc model.ConfigStoreCache) []model.Config
 }
+
+type ProviderV2 interface {
+	// ConvertAthenzModelIntoIstioAuthzPolicy returns the Istio Authorization Policy Resource associated with given athenz model,
+	// namespace, service name and service label
+	ConvertAthenzModelIntoIstioAuthzPolicy(athenzModel athenz.Model, namespace string, serviceName string, svcLabel string) model.Config
```
mcieplak, line 25: Any reason we cannot reuse the existing provider interface? I think we should implement ConvertAthenzModelIntoIstioRbac and GetCurrentIstioRbac functions so that we can compare the authz policies we generated with the existing ones on the cluster. This will allow us to vastly reduce the number of api calls we make. Similar to here: https://github.com/yahoo/k8s-athenz-istio-auth/blob/master/pkg/controller/controller.go#L175-L180

pkg/istio/rbac/v2/provider.go
```diff
@@ -0,0 +1,136 @@
...
+				continue
+			}
+			// if svc match with current svc, process it and add it to the rules
+			// note that svc defined on athenz can be a regex, need to match the pattern
+			res, e := regexp.MatchString(svc, svcLabel)
+			if e != nil {
+				log.Errorln("error matching string: ", e.Error())
+				continue
+			}
+			if res {
```
mcieplak, line 65: Should we check for !res here and continue directly? This way we can avoid indenting the subsequent logic.

mcieplak (changes requested):

mdhume (changes requested):

main.go
```diff
@@ -69,6 +70,8 @@ func main() {
 		log.Panicf("Error creating athenz domain client: %s", err.Error())
 	}
 
+	istioClientSet, err := versionedclient.NewForConfig(config)
```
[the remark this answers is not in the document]
RuiGuo34: I believe the type is different, one is `Client`, the other is `ClientSet`

pkg/controller/controller.go
```diff
@@ -213,21 +218,26 @@ func NewController(dnsSuffix string, istioClient *crd.Client, k8sClient kubernet
...
 		adResyncInterval:     adResyncInterval,
 	}
 
-	configStoreCache.RegisterEventHandler(model.ServiceRole.Type, c.processConfigEvent)
-	configStoreCache.RegisterEventHandler(model.ServiceRoleBinding.Type, c.processConfigEvent)
-	configStoreCache.RegisterEventHandler(model.ClusterRbacConfig.Type, crcController.EventHandler)
+	configStoreCache.RegisterEventHandler(collections.IstioRbacV1Alpha1Serviceroles.Resource().GroupVersionKind(), c.processConfigEvent)
+	configStoreCache.RegisterEventHandler(collections.IstioRbacV1Alpha1Servicerolebindings.Resource().GroupVersionKind(), c.processConfigEvent)
+	configStoreCache.RegisterEventHandler(collections.IstioRbacV1Alpha1Clusterrbacconfigs.Resource().GroupVersionKind(), crcController.EventHandler)
+	configStoreCache.RegisterEventHandler(collections.IstioSecurityV1Beta1Authorizationpolicies.Resource().GroupVersionKind(), c.processConfigEvent)
```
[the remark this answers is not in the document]
RuiGuo34: updated

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+
+// sync is responsible for invoking the appropriate API operation on the model.Config resource
+func (c *Controller) sync(item interface{}) error {
+	castItem, ok := item.(Item)
+	if !ok {
+		return fmt.Errorf("Unable to cast interface")
+	}
+
+	// check if resource can be cast to service, athenzdomain, or authorizationpolicy object
+	_, svcCast := (castItem.Resource).(*corev1.Service)
```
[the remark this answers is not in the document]
RuiGuo34: yes, updated

go.mod
```diff
@@ -4,18 +4,29 @@ go 1.12
...
-	k8s.io/apimachinery v0.0.0-20190817221809-bf4de9df677c
-	k8s.io/client-go v10.0.0+incompatible
-	k8s.io/klog v0.4.0 // indirect
-	k8s.io/kube-openapi v0.0.0-20190816220812-743ec37842bf // indirect
+	istio.io/api v0.0.0-20200513175333-ae3da0d240e3
+	istio.io/client-go v0.0.0-20200513180646-f8d9d8ff84e6
+	istio.io/istio v0.0.0-20200708165503-80f49905d910
+	istio.io/pkg v0.0.0-20200324191837-25e6bb9cf135
+	k8s.io/api v0.18.0
+	k8s.io/apimachinery v0.18.0
```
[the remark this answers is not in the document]
RuiGuo34: these versions are pulled in by istio.io/istio package, istio package uses 1.5.8 version

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+		if castItem.Operation == model.EventAdd {
+			// creation of service will check if istio annotation is set to true, will result in creation of authz policy
+			// authz policy creation logic
+			// check if current service has istio authz annotation set
+			if _, ok = obj.Annotations[authzEnabledAnnotation]; ok && c.checkAuthzEnabledAnnotation(obj) {
+				// here should check if authz policy is existing in the cluster
+				log.Infof("istio authz annotation for service %s is set to true\n", obj.Name)
+				err := c.createAuthzPolicyResource(convertedCR, obj)
+				if err != nil {
+					log.Errorln("error creating authz policy: ", err.Error())
```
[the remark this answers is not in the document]
RuiGuo34: it does create an extra space, removed the space

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+	case svcCast:
+		obj, _ := (castItem.Resource).(*corev1.Service)
+		var convertedCR model.Config
+		if castItem.Operation == model.EventAdd {
+			// creation of service will check if istio annotation is set to true, will result in creation of authz policy
+			// authz policy creation logic
+			// check if current service has istio authz annotation set
+			if _, ok = obj.Annotations[authzEnabledAnnotation]; ok && c.checkAuthzEnabledAnnotation(obj) {
+				// here should check if authz policy is existing in the cluster
+				log.Infof("istio authz annotation for service %s is set to true\n", obj.Name)
```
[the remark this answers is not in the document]
RuiGuo34: Infof doesn't create new line, Infoln creates new line itself

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+				err := c.createAuthzPolicyResource(convertedCR, obj)
```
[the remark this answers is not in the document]
RuiGuo34: dry run is handled in func createAuthzPolicyResource()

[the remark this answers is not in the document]
RuiGuo34: dry run logic is in func createAuthzPolicyResource()

pkg/istio/authorizationpolicy/controller.go
RuiGuo34: Sure, @CatherineZhuang, after matching with wildcard services, we can update those policies that have config change

pkg/istio/rbac/interface.go
RuiGuo34: added the function. but computeChangeList will be different, we cannot tell only from athenz domain how many authz policies created, maybe looping through authz files, regenerate and compare, and update those needed. Maybe we can add this logic in a later pr because we are enabling dry run only this time.

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,463 @@
...
+					log.Infoln("Revision number is: ", revision)
+				} else {
+					err := c.createDryrunResource(convertedCR, authzPolicy.Name, obj.Namespace)
+					if err != nil {
+						return fmt.Errorf("unable write to file, err: %v", err)
+					}
+				}
+			}
+		}
+	case apCast:
```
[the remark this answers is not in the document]
RuiGuo34: they can create with annotation "overrideAuthzPolicy", which won't be controlled by controller.

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,456 @@
...
+	}
+
+	athenzDomain, ok := athenzDomainRaw.(*adv1.AthenzDomain)
+	if !ok {
+		return errors.New("athenz domain cast failed")
+	}
+	signedDomain := athenzDomain.Spec.SignedDomain
+	labels := obj.GetLabels()
+	domainRBAC := athenz.ConvertAthenzPoliciesIntoRbacModel(signedDomain.Domain, &c.adIndexInformer)
+	if _, ok := labels["svc"]; !ok {
```
CatherineZhuang, line 270: Should we move the labels check under `labels := obj.GetLabels()`? 

This code block
```domainRBAC := athenz.ConvertAthenzPoliciesIntoRbacModel(signedDomain.Domain, &c.adIndexInformer)
convertedCR = c.rbacProvider.ConvertAthenzModelIntoIstioAuthzPolicy(domainRBAC, obj.Namespace, obj.Name, labels["svc"])
log.Infoln("Creating Authz Policy ... ")
if !c.dryrun {
	revision, err := c.configStoreCache.Create(convertedCR)
	if err != nil {
		log.Errorln("error creating authz policy:", err.Error())
		return err
	}
	log.Debugln("Created revision number is: ", revision)
} else {
	err := c.createDryrunResource(convertedCR, obj.Name, obj.Namespace)
	if err != nil {
		return fmt.Errorf("unable write to file, err: %v", err)
	}
}
```
seems to be used multiple times for create/update in processSvcResource/processAthenzDomainResource/processAuthorizationPolicyResource. Can we make a function and reuse it? We can add if statements inside the function to check for create/update/delete operations maybe?

pkg/istio/rbac/v1/provider.go
```diff
@@ -96,15 +101,160 @@ func (p *v1) ConvertAthenzModelIntoIstioRbac(m athenz.Model) []model.Config {
...
+	out.Spec = spec
+	return out
+}
+
+// GetCurrentIstioRbac returns the authorization policies resources for the specified model's namespace
+func (p *v1) GetCurrentIstioAuthzPolicy(m athenz.Model, csc model.ConfigStoreCache) []model.Config {
+
+	ap, err := csc.List(collections.IstioSecurityV1Beta1Authorizationpolicies.Resource().GroupVersionKind(), m.Namespace)
+	if err != nil {
+		log.Errorf("Error listing the ServiceRole resources in the namespace: %s", m.Namespace)
```
CatherineZhuang, line 256: minor: should this be "Error listing AuthorizatonPolicy resources?"

pkg/istio/authorizationpolicy/controller.go
RuiGuo34: I think for events, we can keep it separate for better readability. I updated conversion into a help function to reduce code duplication

pkg/istio/rbac/v1/provider.go
RuiGuo34: fixed

pkg/istio/rbac/v1/provider.go
```diff
@@ -96,15 +101,167 @@ func (p *v1) ConvertAthenzModelIntoIstioRbac(m athenz.Model) []model.Config {
...
+		for _, assert := range assertions {
+			// assert.Resource contains the svc information that needs to parse and match
+			svc, path, err := common.ParseAssertionResource(zms.DomainName(athenz.NamespaceToDomain(namespace)), assert)
+			if err != nil {
+				continue
+			}
+			// if svc match with current svc, process it and add it to the rules
+			// note that svc defined on athenz can be a regex, need to match the pattern
+			res, e := regexp.MatchString(svc, svcLabel)
+			if e != nil {
```
mcieplak, line 154: Can we call this variable err?

pkg/istio/rbac/v1/provider.go
```diff
@@ -96,15 +101,167 @@ func (p *v1) ConvertAthenzModelIntoIstioRbac(m athenz.Model) []model.Config {
...
+				log.Errorln("error matching string: ", e.Error())
+				continue
+			}
+			if !res {
+				log.Errorf("athenz svc %s does not match with current svc %s", svc, svcLabel)
```
mcieplak, line 159: Wondering if we should make this debug, this might spam for athenz domains with a lot of policies defined?

pkg/istio/rbac/v1/provider.go
```diff
@@ -96,15 +101,167 @@ func (p *v1) ConvertAthenzModelIntoIstioRbac(m athenz.Model) []model.Config {
...
+			from_principal := &v1beta1.Rule_From{
+				Source: &v1beta1.Source{},
+			}
+			from_requestPrincipal := &v1beta1.Rule_From{
+				Source: &v1beta1.Source{},
+			}
+			// role name should match zms resource name
+			for _, roleName := range athenzModel.Roles {
+				if roleName == role {
+					// add function to enableOriginJwtSubject, following code assume enableOriginJwtSubject is true by default
```
mcieplak, line 173: Should we remove this comment? Looks like we are doing this check.

pkg/istio/rbac/v1/provider.go
```diff
@@ -96,15 +101,167 @@ func (p *v1) ConvertAthenzModelIntoIstioRbac(m athenz.Model) []model.Config {
...
+						}
+					}
+					//add role spiffee for role certificate
+					roleSpiffeName, err := common.RoleToSpiffe(string(athenzModel.Name), string(roleName))
+					if err != nil {
+						log.Println("error when convert role to spiffe name: ", err.Error())
+						continue
+					}
+					from_principal.Source.Principals = append(from_principal.Source.Principals, roleSpiffeName)
+				}
```
mcieplak, line 216: Should we break from the loop here after we found the role to avoid additional iterations?

pkg/istio/rbac/v1/provider.go
```diff
@@ -96,15 +101,167 @@ func (p *v1) ConvertAthenzModelIntoIstioRbac(m athenz.Model) []model.Config {
...
+			// form rule.From, must initialize internal source here
+			from_principal := &v1beta1.Rule_From{
+				Source: &v1beta1.Source{},
+			}
+			from_requestPrincipal := &v1beta1.Rule_From{
+				Source: &v1beta1.Source{},
+			}
+			// role name should match zms resource name
+			for _, roleName := range athenzModel.Roles {
+				if roleName == role {
```
mcieplak, line 172: Can't we skip this outer for loop and if check and just directly use the `for _, roleMember := range athenzModel.Members[roleName]` loop instead? We already have the role here so we could use `role` instead of `roleName` to do the lookup against the map?

pkg/istio/rbac/v1/provider.go
```diff
@@ -96,15 +101,167 @@ func (p *v1) ConvertAthenzModelIntoIstioRbac(m athenz.Model) []model.Config {
...
+	}
+	spec.Rules = rules
+	out.Spec = spec
+	return out
+}
+
+// GetCurrentIstioRbac returns the authorization policies resources for the specified model's namespace
+// if serviceName is "", return the all the authorization policies in the given namespace,
+// if serviceName is specific, return single authorization policy matching with serviceName.
+func (p *v1) GetCurrentIstioAuthzPolicy(m athenz.Model, csc model.ConfigStoreCache, serviceName string) []model.Config {
```
mcieplak, line 254: Could we use the GetCurrentIstioRbac function from the interface? We could update it to pass in the service name.

pkg/istio/rbac/v1/provider.go
```diff
@@ -96,15 +101,167 @@ func (p *v1) ConvertAthenzModelIntoIstioRbac(m athenz.Model) []model.Config {
...
-	srb, err := csc.List(model.ServiceRoleBinding.Type, m.Namespace)
+	srb, err := csc.List(collections.IstioRbacV1Alpha1Servicerolebindings.Resource().GroupVersionKind(), m.Namespace)
 	if err != nil {
 		log.Errorf("Error listing the ServiceRoleBinding resources in the namespace: %s", m.Namespace)
 	}
 
 	return append(sr, srb...)
 }
+
+// ConvertAthenzModelIntoIstioAuthzPolicy converts the Athenz RBAC model into Istio Authorization V1Beta1 specific
```
mcieplak, line 117: Is it possible to modify the `ConvertAthenzModelIntoIstioRbac` function to work with `ConvertAthenzModelIntoIstioAuthzPolicy` so that we can keep the same interface function?

pkg/istio/rbac/v1/provider_test.go
```diff
@@ -868,3 +874,158 @@ func TestGetCurrentIstioRbac(t *testing.T) {
...
+				"authz.istio.io/enabled": "true",
+			},
+			Labels: map[string]string{
+				"app": "productpage",
+			},
+		},
+	}
+)
+
+func TestConvertAthenzModelIntoIstioAuthzPolicy(t *testing.T) {
```
mcieplak, line 899: We should test the same functionality we're currently testing in the service role / service role binding tests. I'm okay if we make this in a separate PR.

pkg/istio/rbac/v1/provider.go
```diff
@@ -96,15 +101,167 @@ func (p *v1) ConvertAthenzModelIntoIstioRbac(m athenz.Model) []model.Config {
...
+	spec.Selector = &workloadv1beta1.WorkloadSelector{
+		MatchLabels: map[string]string{"svc": svcLabel},
+	}
+
+	// generating rules, iterate through assertions, find the one match with desired format.
+	var rules []*v1beta1.Rule
+	for role, assertions := range athenzModel.Rules {
+		for _, assert := range assertions {
+			// assert.Resource contains the svc information that needs to parse and match
+			svc, path, err := common.ParseAssertionResource(zms.DomainName(athenz.NamespaceToDomain(namespace)), assert)
```
mcieplak, line 147: We can use the domain name directly from the athenz model here.

pkg/istio/rbac/v1/provider.go
```diff
@@ -96,15 +101,167 @@ func (p *v1) ConvertAthenzModelIntoIstioRbac(m athenz.Model) []model.Config {
...
+	var out model.Config
+	// form authorization config meta
+	// namespace: service's namespace
+	// name: service's name
+	schema := collections.IstioSecurityV1Beta1Authorizationpolicies
+	out.ConfigMeta = model.ConfigMeta{
+		Type:      schema.Resource().Kind(),
+		Group:     schema.Resource().Group(),
+		Version:   schema.Resource().Version(),
+		Namespace: namespace,
```
mcieplak, line 131: We could potentially skip passing the namespace as an argument as we have the athenz domain name in the athenz model object.

mcieplak (changes requested):

go.mod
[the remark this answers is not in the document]
mcieplak: That makes sense but is there a reason the version doesn't match the replace? For example, we have `k8s.io/apimachinery v0.18.0` but below we have `replace k8s.io/apimachinery => k8s.io/apimachinery v0.17.14`. Shouldn't both of these be the same version?

pkg/istio/rbac/v1/provider.go
```diff
@@ -96,15 +101,167 @@ func (p *v1) ConvertAthenzModelIntoIstioRbac(m athenz.Model) []model.Config {
...
+		if err != nil {
+			log.Errorf("Error listing the Authorization Policy resources in the namespace: %s", m.Namespace)
+		}
+		return apList
+	}
+	ap := csc.Get(collections.IstioSecurityV1Beta1Authorizationpolicies.Resource().GroupVersionKind(), serviceName, m.Namespace)
+	if ap != nil {
+		return []model.Config{*ap}
+	}
+	return []model.Config{}
```
mcieplak, line 266: Don't we need to support the dry run mode here too and read from the files in the pod?

pkg/istio/onboarding/controller_test.go
```diff
@@ -295,17 +294,19 @@ func TestGetServiceList(t *testing.T) {
...
+			Group:   collections.IstioRbacV1Alpha1Clusterrbacconfigs.Resource().Group(),
+			Version: collections.IstioRbacV1Alpha1Clusterrbacconfigs.Resource().Version(),
 		},
 		Spec: &v1alpha1.RbacConfig{
 			Mode: v1alpha1.RbacConfig_ON_WITH_EXCLUSION,
 			Exclusion: &v1alpha1.RbacConfig_Target{
 				Services: services,
 			},
-			Inclusion: nil,
+			Inclusion: &v1alpha1.RbacConfig_Target{
```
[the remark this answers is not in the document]
RuiGuo34: I think in the istio library upgrade, Inclusion field does not support nil anymore, test was failing

pkg/istio/onboarding/controller.go
```diff
@@ -277,11 +279,12 @@ func (c *Controller) sync() error {
 			ConfigMeta: config.ConfigMeta,
 			Spec:       clusterRbacConfig,
 		}
-		item := processor.Item{
+		item := common.Item{
 			Operation:       model.EventUpdate,
 			Resource:        config,
 			CallbackHandler: c.callbackHandler,
 		}
+		// TODO: investigate: when dns-suffix is changed and results in full service list update, this update is likely to fail and controller will be stuck in fail and retry cycles.
```
mdhume, line 287: can you elaborate on this issue

mdhume (changes requested):

main.go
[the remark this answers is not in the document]
RuiGuo34: client internal uses this, I need to define this in order to initialize the client

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,438 @@
...
+	"strings"
+	"time"
+)
+
+const (
+	queueNumRetries            = 3
+	authzEnabled               = "true"
+	authzEnabledAnnotation     = "authz.istio.io/enabled"
+	overrideAnnotation         = "overrideAuthzPolicy"
+	DryRunStoredFilesDirectory = "/root/authzpolicy/"
```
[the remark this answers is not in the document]
RuiGuo34: no, I will change it to lower case

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,438 @@
...
+	adIndexInformer          cache.SharedIndexInformer
+	authzpolicyIndexInformer cache.SharedIndexInformer
+	queue                    workqueue.RateLimitingInterface
+	rbacProvider             rbac.Provider
+	apResyncInterval         time.Duration
+	enableOriginJwtSubject   bool
+	dryrun                   bool
+}
+
+func NewController(configStoreCache model.ConfigStoreCache, serviceIndexInformer cache.SharedIndexInformer, adIndexInformer cache.SharedIndexInformer, istioClientSet versioned.Interface, apResyncInterval time.Duration, enableOriginJwtSubject bool, DryRun bool) *Controller {
```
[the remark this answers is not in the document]
RuiGuo34: updated

pkg/istio/rbac/v1/provider.go
```diff
@@ -96,15 +101,167 @@ func (p *v1) ConvertAthenzModelIntoIstioRbac(m athenz.Model) []model.Config {
...
+						continue
+					}
+					from_principal.Source.Principals = append(from_principal.Source.Principals, roleSpiffeName)
+				}
+			}
+			rule.From = append(rule.From, from_principal)
+			rule.From = append(rule.From, from_requestPrincipal)
+			// form rules_to
+			rule_to := &v1beta1.Rule{}
+			_, err = common.ParseAssertionEffect(assert)
```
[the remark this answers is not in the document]
RuiGuo34: action in authz policy is at top level, which applies to all the rules. ParseAssertionEffect results is for single policy. I think we can think about adding it when it comes to DENY/ALLOW action implementation.

pkg/istio/rbac/common/util.go
```diff
@@ -69,3 +272,151 @@ func NewConfig(configType string, namespace string, name string, spec proto.Mess
...
+// findDeleteDryrunResource retrieves the yaml file from local directory and deletes it
+func (d *DryRunHandler) findDeleteDryrunResource(item *Item) error {
+	authzPolicyName := item.Resource.ConfigMeta.Name
+	namespace := item.Resource.ConfigMeta.Namespace
+	yamlFileName := authzPolicyName + "--" + namespace + ".yaml"
+	if _, err := os.Stat(dryRunStoredFilesDirectory + yamlFileName); os.IsNotExist(err) {
+		log.Infof("file %s does not exist in local directory", dryRunStoredFilesDirectory+yamlFileName)
+		return nil
+	} else if err != nil {
+		log.Infof("error stating file %s in local directory: %s, error: %s", dryRunStoredFilesDirectory+yamlFileName, err)
```
mcieplak, line 380: Should we return the error here?

pkg/istio/rbac/common/util.go
```diff
@@ -69,3 +272,151 @@ func NewConfig(configType string, namespace string, name string, spec proto.Mess
...
+	// define istio object interface to unmarshall yaml object into
+	item := &crd.IstioKind{Spec: map[string]interface{}{}}
+	yamlFileName := serviceName + "--" + namespace + ".yaml"
+	yamlFile, err := ioutil.ReadFile(dryRunStoredFilesDirectory + yamlFileName)
+	if err != nil {
+		return nil, fmt.Errorf("unable to read yaml file to local directory, err: %s", err)
+	}
+	err = yaml.Unmarshal(yamlFile, item)
+	if err != nil {
+		return nil, fmt.Errorf("unable to unmarshall yaml file, err: %s", err)
```
mcieplak, line 397: Typo: unmarshal.

pkg/istio/rbac/common/util.go
```diff
@@ -69,3 +272,151 @@ func NewConfig(configType string, namespace string, name string, spec proto.Mess
...
+	config, err := crd.ConvertObject(collections.IstioSecurityV1Beta1Authorizationpolicies, item, "")
+	if err != nil {
+		return nil, fmt.Errorf("unable to convert yaml converted istio object to authorization policy model config, err: %s", err)
+	}
+	return config, nil
+}
+
+// FetchServices walks through the files under a directory, based on the naming convention, it parses the service name
+// from the file name.
+func FetchServices(namespace string) []string {
```
mcieplak, line 408: Nit: could we rename this `FetchServicesFromDir`?

pkg/istio/rbac/common/util.go
```diff
@@ -69,3 +272,151 @@ func NewConfig(configType string, namespace string, name string, spec proto.Mess
...
+		return nil
+	} else if err != nil {
+		log.Infof("error stating file %s in local directory: %s, error: %s", dryRunStoredFilesDirectory+yamlFileName, err)
+	}
+	log.Infof("deleting file under path: %s\n", dryRunStoredFilesDirectory+yamlFileName)
+	return os.Remove(dryRunStoredFilesDirectory + yamlFileName)
+}
+
+// ReadConvertToModelConfig reads in the authorization policy yaml object and converts it into a model.Config struct
+func ReadConvertToModelConfig(serviceName, namespace string) (*model.Config, error) {
```
mcieplak, line 387: Can we add simple unit tests for the functions createDryrunResource, findDeleteDryrunResource, ReadConvertToModelConfig, and FetchServices? We can use temp dir to interact with files: https://golang.org/pkg/os/#TempDir

pkg/istio/rbac/interface.go
```diff
@@ -13,8 +13,8 @@ type Provider interface {
 
 	// ConvertAthenzModelIntoIstioRbac converts the given Athenz model into a list of Istio type RBAC resources
 	// Any implementation should return exactly the same list of output resources for a given Athenz model
-	ConvertAthenzModelIntoIstioRbac(model athenz.Model) []model.Config
+	ConvertAthenzModelIntoIstioRbac(athenzModel athenz.Model, serviceName string, svcLabel string) []model.Config
 
 	// GetCurrentIstioRbac returns the Istio RBAC custom resources associated with the given model
-	GetCurrentIstioRbac(model athenz.Model, csc model.ConfigStoreCache) []model.Config
+	GetCurrentIstioRbac(model athenz.Model, csc model.ConfigStoreCache, serviceName string, dryRun bool) []model.Config
```
mcieplak, line 19: Can we make the dryRun flag into the v2 provider object instead of making it an arg in the interface? Similar to the enableOriginJwtSubject flag.

pkg/istio/rbac/v1/provider.go
```diff
@@ -27,7 +27,7 @@ func NewProvider(enableOriginJwtSubject bool) rbac.Provider {
 // ConvertAthenzModelIntoIstioRbac converts the Athenz RBAC model into the list of Istio Authorization V1 specific
 // RBAC custom resources (ServiceRoles, ServiceRoleBindings)
 // The idea is that with a given input model, the function should always return the same output list of resources
-func (p *v1) ConvertAthenzModelIntoIstioRbac(m athenz.Model) []model.Config {
+func (p *v1) ConvertAthenzModelIntoIstioRbac(m athenz.Model, serviceName string, svcLabel string) []model.Config {
```
mcieplak, line 30: Nit: we can make the args we're not using underscores.

pkg/istio/rbac/v1/provider.go
```diff
@@ -86,22 +86,22 @@ func (p *v1) ConvertAthenzModelIntoIstioRbac(m athenz.Model) []model.Config {
...
+		srb := common.NewConfig(collections.IstioRbacV1Alpha1Servicerolebindings, m.Namespace, k8sRoleName, srbSpec)
 		out = append(out, srb)
 	}
 
 	return out
 }
 
 // GetCurrentIstioRbac returns the ServiceRole and ServiceRoleBinding resources for the specified model's namespace
-func (p *v1) GetCurrentIstioRbac(m athenz.Model, csc model.ConfigStoreCache) []model.Config {
+func (p *v1) GetCurrentIstioRbac(m athenz.Model, csc model.ConfigStoreCache, serviceName string, dryrun bool) []model.Config {
```
mcieplak, line 97: Nit: we can make the args we're not using underscores.

pkg/istio/rbac/v2/provider.go
```diff
@@ -0,0 +1,195 @@
...
+		apList, err := csc.List(collections.IstioSecurityV1Beta1Authorizationpolicies.Resource().GroupVersionKind(), namespace)
+		if err != nil {
+			log.Errorf("Error listing the Authorization Policy resources in the namespace: %s", namespace)
+		}
+		return apList
+	}
+	ap := csc.Get(collections.IstioSecurityV1Beta1Authorizationpolicies.Resource().GroupVersionKind(), serviceName, namespace)
+	if ap != nil {
+		return []model.Config{*ap}
+	}
```
mcieplak, line 193: Should we add a log statement for the else here in case it is nil?

pkg/istio/rbac/v2/provider_test.go
```diff
@@ -0,0 +1,175 @@
...
+	"istio.io/istio/pkg/config/schema/collections"
+	k8sv1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/tools/cache"
+	"testing"
+	"time"
+)
+
+const (
+	DomainName = "test.namespace"
```
mcieplak, line 25: Should we make this private?

pkg/istio/rbac/v2/provider_test.go
```diff
@@ -0,0 +1,175 @@
...
+	username   = "user.name"
+	username1  = "user.*"
```
mcieplak, line 27: Nit: can we rename this to wildcardUsername?

pkg/istio/rbac/v2/provider_test.go
```diff
@@ -0,0 +1,175 @@
...
+				"authz.istio.io/enabled": "true",
+			},
+			Labels: map[string]string{
+				"app": "productpage",
+			},
+		},
+	}
+)
+
+func TestConvertAthenzModelIntoIstioRbac(t *testing.T) {
```
mcieplak, line 45: Can we add more unit tests here to cover the same cases we're covering for the provider v1 flow in a subsequent PR?

mcieplak (changes requested):

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,360 @@
...
+	}
+}
+
+// ProcessConfigChange receives resource and event action, and perform update on resource
+func (c *Controller) processConfigChange(item *common.Item) error {
+	if item == nil {
+		return nil
+	}
+	var err error
+	var eHandler common.EventHandler
```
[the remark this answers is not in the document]
mcieplak: Ah nevermind, I see that this will need to change based on if the dry run is enabled per service. Does it make sense to create both the dry run and api handler objects in the controller then and then call the corresponding one based on if it's in dry run mode or not for that particular service?

pkg/istio/rbac/v2/provider_test.go
RuiGuo34: will do, having a pending pr for it https://github.com/yahoo/k8s-athenz-istio-auth/pull/48

pkg/controller/controller.go
```diff
@@ -256,7 +192,7 @@ func (c *Controller) processEvent(fn cache.KeyFunc, obj interface{}) {
 }
 
 // processConfigEvent is responsible for adding the key of the item to the queue
-func (c *Controller) processConfigEvent(config model.Config, e model.Event) {
+func (c *Controller) processConfigEvent(config model.Config, _ model.Config, e model.Event) {
```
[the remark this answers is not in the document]
RuiGuo34: need to update this function. Basically, first parameter is the old Config, when you try to update an resource, this config will be the previous applied config, second parameter is the current you are going to apply. When there ADD, DELETE event, the first field would be empty.

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -0,0 +1,360 @@
...
+
+	signedDomain := athenzDomain.Spec.SignedDomain
+	domainRBAC := athenz.ConvertAthenzPoliciesIntoRbacModel(signedDomain.Domain, &c.adIndexInformer)
+	var serviceList []*corev1.Service
+	if serviceName != "" {
+		serviceObj, err := c.getSvcObj(athenz.DomainToNamespace(athenzDomainName) + "/" + serviceName)
+		if err != nil {
+			return fmt.Errorf("error getting service from cache: %s", err.Error())
+		}
+		if serviceObj != nil {
```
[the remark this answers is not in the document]
RuiGuo34: when service doesn't exist in the cache, updated the getSvcObj function. Basically when creating service, it should be nil from cache, and controller with try to create authz policy config

pkg/istio/rbac/common/util.go
```diff
@@ -5,25 +5,136 @@ package common
...
-	roleName := strings.TrimPrefix(roleFQDN, fmt.Sprintf("%s:role.", domainName))
-	if strings.Contains(roleName, ":") {
-		return "", fmt.Errorf("role: %s does not belong to the Athenz domain: %s", roleFQDN, domainName)
+const (
+	allUsers                     = "user.*"
+	WildCardAll                  = "*"
+	ServiceRoleKind              = "ServiceRole"
+	AthenzJwtPrefix              = "athenz/"
+	RequestAuthPrincipalProperty = "request.auth.principal"
+	dryRunStoredFilesDirectory   = "/root/authzpolicy/"
```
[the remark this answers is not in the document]
RuiGuo34: I think importing from main doesn't look good: https://stackoverflow.com/questions/35038864/how-to-access-global-variables. import from main package will also cause circular dependency

pkg/istio/rbac/v1/provider.go
RuiGuo34: for this function, I did not see any variable has underscores

pkg/istio/rbac/common/util.go
```diff
@@ -69,3 +272,152 @@ func NewConfig(configType string, namespace string, name string, spec proto.Mess
...
+	}
+	return config, nil
+}
+
+// FetchServicesFromDir walks through the files under a directory, based on the naming convention, it parses the service name
+// from the file name.
+func FetchServicesFromDir(namespace, localDirPath string) []string {
+	files, err := ioutil.ReadDir(localDirPath)
+	if err != nil {
+		log.Errorf("unable to read yaml files to local directory, err: %s", err)
```
mcieplak, line 412: Should we return an error here?

pkg/istio/rbac/common/util.go
```diff
@@ -69,3 +272,152 @@ func NewConfig(configType string, namespace string, name string, spec proto.Mess
...
+	} else if err != nil {
+		return fmt.Errorf("error stating file %s in local directory, error: %s", localDirPath+yamlFileName, err)
+	}
+	log.Infof("deleting file under path: %s\n", localDirPath+yamlFileName)
+	return os.Remove(localDirPath + yamlFileName)
+}
+
+// ReadConvertToModelConfig reads in the authorization policy yaml object and converts it into a model.Config struct
+func ReadConvertToModelConfig(serviceName, namespace, localDirPath string) (*model.Config, error) {
+	// define istio object interface to unmarshall yaml object into
```
mcieplak, line 389: Typo: unmarshal

pkg/istio/rbac/common/util_test.go
```diff
@@ -239,3 +535,484 @@ func TestNewConfig(t *testing.T) {
...
+		}
+
+		// convert the created yaml back to config model format, compare the model spec
+		covertedConfig, err := ReadConvertToModelConfig(tt.item.Resource.Name, tt.item.Resource.Namespace, os.TempDir()+"/")
+		assert.Equal(t, tt.expErr, err, "error should be nil when converting config")
+		assert.Equal(t, *covertedConfig, tt.item.Resource, "model config should be the same")
+
+		// delete the created resource
+		err = eHandler.findDeleteDryrunResource(&tt.item, os.TempDir()+"/")
+		assert.Equal(t, tt.expErr, err, "error should not be nil when deleting the yaml file")
```
mcieplak, line 962: Can we also add a stat check here to make sure the file does not exist?

pkg/istio/rbac/common/util_test.go
```diff
@@ -239,3 +535,484 @@ func TestNewConfig(t *testing.T) {
...
+		{
+			item:            getAuthzPolicyItem(model.EventAdd),
+			fileName:        "onboarded-service--test-namespace.yaml",
+			expectedContent: "onboarded-service--test-namespace.yaml",
+			expErr:          nil,
+		},
+	}
+
+	for _, tt := range tests {
+		// test findDeleteDryrunResource func
```
mcieplak, line 947: Can we update the comment here?

pkg/istio/rbac/common/util_test.go
```diff
@@ -239,3 +535,484 @@ func TestNewConfig(t *testing.T) {
...
+		})
+	}
+}
+
+func TestDryrunResource(t *testing.T) {
+	eHandler := DryRunHandler{}
+	tests := []struct {
+		item            Item
+		fileName        string
+		expectedContent string
```
mcieplak, line 935: I believe we can get rid of the expectedContent variable here.

pkg/istio/rbac/v2/provider.go
```diff
@@ -0,0 +1,204 @@
...
+	if serviceName == "" {
+		apList, err := csc.List(collections.IstioSecurityV1Beta1Authorizationpolicies.Resource().GroupVersionKind(), namespace)
+		if err != nil {
+			log.Errorf("Error listing the Authorization Policy resources in the namespace: %s", namespace)
+		}
+		return apList
+	}
+	ap := csc.Get(collections.IstioSecurityV1Beta1Authorizationpolicies.Resource().GroupVersionKind(), serviceName, namespace)
+	if ap != nil {
+		log.Infof("authorization policy does not exist in the cache, name: %s, namespace: %s", serviceName, namespace)
```
mcieplak, line 200: I believe this should be part of the else statement here.

pkg/istio/rbac/v2/provider.go
```diff
@@ -147,8 +96,65 @@ func (p *v2) ConvertAthenzModelIntoIstioRbac(athenzModel athenz.Model, serviceNa
...
+		}
+		//add role spiffe for role certificate
+		roleSpiffeName, err := common.RoleToSpiffe(string(athenzModel.Name), string(role))
+		if err != nil {
+			log.Errorln("error when convert role to spiffe name: ", err.Error())
+			continue
+		}
+		from_principal.Source.Principals = append(from_principal.Source.Principals, roleSpiffeName)
+		rule.From = append(rule.From, from_principal)
+		rule.From = append(rule.From, from_requestPrincipal)
```
mcieplak, line 156: Should we append this only if enableOriginJwtSubject is enabled? Not sure how an empty array will be treated here if the flag is off.

mcieplak (changes requested):

pkg/istio/rbac/v2/provider.go
RuiGuo34: makes sense

mcieplak (approved):

mdhume (approved):

PR #48, opened 2021-02-04
update authz policy test cases

RuiGuo34:
@mcieplak @mdhume please review


mcieplak (changes requested):

pkg/istio/authorizationpolicy/controller_test.go
```diff
@@ -136,17 +166,276 @@ func newFakeController(services []*v1.Service, fake bool, stopCh <-chan struct{}
...
+
+func TestSyncAthenzDomain(t *testing.T) {
+	tests := []struct {
+		name                string
+		expectedAuthzPolicy model.Config
+		fake                bool
+		item                Item
+	}{
+		{
+			name:                "update existing authz policy spec when there is athenz domain crd update",
```
mcieplak, line 292: Can we also add a test case to check if athenz domain create and delete are working correctly?

pkg/istio/authorizationpolicy/controller_test.go
```diff
@@ -26,48 +30,75 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"sync"
 	"testing"
+	"time"
 )
 
+type Item struct {
```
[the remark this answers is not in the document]
RuiGuo34: not sure how to directly use it, item in common library has interface type, but here I want to use model.Config type

pkg/istio/authorizationpolicy/controller_test.go
```diff
@@ -285,47 +281,80 @@ func TestSyncAthenzDomain(t *testing.T) {
...
 	}{
 		{
 			name:                "update existing authz policy spec when there is athenz domain crd update",
-			expectedAuthzPolicy: getExpectedCR(),
-			fake:                true,
+			expectedAuthzPolicy: getExpectedAuthzPolicy(),
 			item:                Item{Operation: model.EventUpdate, Resource: onboardedAthenzDomain},
 		},
+		{
+			name:                "no action on authz policy when athenz doamin crd is deleted",
```
mcieplak, line 293: Typo in word domain.

pkg/istio/authorizationpolicy/controller_test.go
```diff
@@ -378,55 +402,47 @@ func TestSyncAuthzPolicy(t *testing.T) {
...
 			switch action := tt.item.Operation; action {
 			case model.EventUpdate:
-				c := newFakeController(tt.inputAthenzDomain, tt.inputService, tt.fake, make(chan struct{}))
-				resourceVersion, err := c.configStoreCache.Create(getOldCR())
-				if err != nil {
-					log.Panicln("controller not able to create authz policy spec: ", err)
+				c := newFakeController(tt.inputAthenzDomain, tt.inputService, true, make(chan struct{}))
+				authzPolicyConfig := getOldAuthzPolicy()
+				annotations := (tt.item.Resource).(*authz.AuthorizationPolicy).Annotations
+				if annotations != nil {
```
mcieplak, line 408: Why is this annotations check needed here?

pkg/istio/authorizationpolicy/controller_test.go
```diff
@@ -378,55 +402,47 @@ func TestSyncAuthzPolicy(t *testing.T) {
...
+				tt.expectedAuthzPolicy.ConfigMeta.ResourceVersion = genAuthzPolicy.ConfigMeta.ResourceVersion
+				assert.Equal(t, tt.expectedAuthzPolicy, *genAuthzPolicy, "created authorization policy spec should be equal")
 			case model.EventDelete:
-				c := newFakeController(&adv1.AthenzDomain{}, &v1.Service{}, tt.fake, make(chan struct{}))
-				c.adIndexInformer.GetStore().Add(tt.inputAthenzDomain.DeepCopy())
-				c.serviceIndexInformer.GetStore().Add(tt.inputService.DeepCopy())
+				c := newFakeController(&adv1.AthenzDomain{}, &v1.Service{}, true, make(chan struct{}))
+				err := c.adIndexInformer.GetStore().Add(tt.inputAthenzDomain.DeepCopy())
+				assert.Nil(t, err, "add athenz domain crd to cache should not return error")
+				err = c.serviceIndexInformer.GetStore().Add(tt.inputService.DeepCopy())
```
mcieplak, line 428: Should we directly pass the athenz domain and service objects into newFakeController?

pkg/istio/authorizationpolicy/controller_test.go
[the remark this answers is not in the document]
mcieplak: Ah I see, the use case is slightly different here too. It's fine to leave this.

mcieplak (changes requested):

mcieplak (approved):

pkg/istio/authorizationpolicy/controller_test.go
```diff
@@ -217,7 +571,95 @@ func getFakeDomain() zms.SignedDomain {
 				{
 					Members:  []zms.MemberName{"productpage-reader"},
 					Modified: &timestamp,
-					Name:     domainName + ":role.productpage-reader",
+					Name:     domainNameOnboarded + ":role.productpage-reader",
```
mdhume, line 574: can we also create a `productpage-writer` role with 2 valid members, so that the created authz policy has a spec with source.Principals other than `*` and role name

mdhume: or better still get it added to the v2 provider unit tests instead

pkg/istio/authorizationpolicy/controller_test.go
```diff
@@ -123,30 +154,300 @@ func newFakeController(services []*v1.Service, fake bool, stopCh <-chan struct{}
...
+			inputAthenzDomain:   onboardedAthenzDomain,
+			initAuthzPolicySpec: getModifiedAuthzPolicyCRWithOverrideAnnotation(),
+			item:                Item{Operation: model.EventUpdate, Resource: getModifiedAuthzPolicyWithOverrideAnnotation()},
+			expectedAuthzPolicy: getModifiedAuthzPolicyCRWithOverrideAnnotation(),
+		},
+		{
+			name:                "when there is manual creation of authz policy without override annotation, controller should delete this create resource",
+			inputService:        notOnboardedServiceWithAnnotation,
+			inputAthenzDomain:   onboardedAthenzDomain,
+			initAuthzPolicySpec: getOldAuthzPolicy(),
```
mdhume, line 400: shouldn't this be set to `model.Config{}` since the authz policy is being manually created

mdhume (changes requested):

pkg/istio/authorizationpolicy/controller_test.go
```diff
@@ -123,30 +154,300 @@ func newFakeController(services []*v1.Service, fake bool, stopCh <-chan struct{}
...
+				err = c.sync(key)
+				assert.Nil(t, err, "sync function should not return error")
+
+				// Updated Spec should be equal
+				genAuthzPolicy := c.configStoreCache.Get(collections.IstioSecurityV1Beta1Authorizationpolicies.Resource().GroupVersionKind(), tt.inputService.Name, tt.inputService.Namespace)
+				if genAuthzPolicy != nil {
+					// set creation timestamp and resource version for expected authz policy resource, which are generated on the fly
+					tt.expectedAuthzPolicy.ConfigMeta.CreationTimestamp = genAuthzPolicy.ConfigMeta.CreationTimestamp
+					tt.expectedAuthzPolicy.ConfigMeta.ResourceVersion = genAuthzPolicy.ConfigMeta.ResourceVersion
+					assert.Equal(t, tt.expectedAuthzPolicy, *genAuthzPolicy, "created authorization policy spec should be equal")
```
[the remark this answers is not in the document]
RuiGuo34: if authz spec exists in the cache, genAuthzPolicy will nil, taking the value of a nil address will get panic error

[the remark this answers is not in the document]
mdhume: got it, should we just assert for nil value then instead? expectedAuthzPolicy should also be nil in that case?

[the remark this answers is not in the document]
RuiGuo34: if expectedAuthzPolicy is type *model.Config, it should work. let me update

pkg/istio/authorizationpolicy/controller_test.go
RuiGuo34: makes sense to add it to the provide unit test

mdhume (approved):

PR #50, opened 2021-02-19
Add command line arguments and edit dry run logic for migration changes 

CatherineZhuang:

Added migration logic:
- enableAuthzPolicyController and authzPolicyEnabledList command-line flags
- edited logic to enable dry run as default and create authzpolicy resources for services in the command line argument


main.go
```diff
@@ -95,7 +96,12 @@ func main() {
 		log.Panicf("Error parsing ap-resync-interval duration: %s", err.Error())
 	}
 
-	c := controller.NewController(*dnsSuffix, istioClient, k8sClient, adClient, istioClientSet, adResyncInterval, crcResyncInterval, apResyncInterval, *enableOriginJwtSubject, *apDryRun)
+	componentsEnabledAuthzPolicy, err := common.ParseComponentsEnabledAuthzPolicy(*authzPolicyEnabledList)
```
mcieplak, line 99: Should we parse this only if the authz policy controller is enabled?

pkg/controller/controller.go
```diff
@@ -146,24 +148,28 @@ func NewController(dnsSuffix string, istioClient *crd.Client, k8sClient kubernet
 	processor := processor.NewController(configStoreCache)
 	crcController := onboarding.NewController(configStoreCache, dnsSuffix, serviceIndexInformer, crcResyncInterval, processor)
 	adIndexInformer := adInformer.NewAthenzDomainInformer(adClient, 0, cache.Indexers{})
-	apController := authzpolicy.NewController(configStoreCache, serviceIndexInformer, adIndexInformer, istioClientSet, apResyncInterval, enableOriginJwtSubject, apDryRun)
+	apController := &authzpolicy.Controller{}
```
mcieplak, line 151: Should we just set this to nil instead by default to minimize memory? Using `var apController *authzpolicy.Controller`

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -232,6 +233,7 @@ func (c *Controller) sync(key string) error {
 	}
 
 	// get current APs from cache
+	// dryRun := !c.componentEnabledAuthzPolicy.IsEnabled(serviceName, domainRBAC.Namespace)
```
mcieplak, line 236: Can we remove this comment?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -55,18 +57,17 @@ func NewController(configStoreCache model.ConfigStoreCache, serviceIndexInformer
 		adIndexInformer:          adIndexInformer,
 		authzpolicyIndexInformer: authzpolicyIndexInformer,
 		queue:                    queue,
-		rbacProvider:             rbacv2.NewProvider(dryRun, enableOriginJwtSubject),
-		apResyncInterval:         apResyncInterval,
-		enableOriginJwtSubject:   enableOriginJwtSubject,
-		dryRun:                   dryRun,
+		// TODO: fix methods in rbacv2 provider funcs, find a way to check dry run flag for current ns / svc
```
mcieplak, line 60: Can we remove this comment now? We do have the dry run check now which will use the corresponding event handler.

pkg/istio/rbac/common/util.go
```diff
@@ -421,3 +421,81 @@ func FetchServicesFromDir(namespace, localDirPath string) ([]string, error) {
...
+	for _, item := range c.serviceList {
+		if item.service == service && item.namespace == ns {
+			return true
+		}
+	}
+	return false
+}
+
+func (c *ComponentEnabled) containsNamespace(ns string) bool {
+	for _, item := range c.namespaceList {
```
mcieplak, line 482: Should we make this a map instead with the key being the namespace so the lookup is O(1)?

pkg/istio/rbac/common/util.go
```diff
@@ -421,3 +421,81 @@ func FetchServicesFromDir(namespace, localDirPath string) ([]string, error) {
...
+			}
+		}
+	}
+	componentEnabledObj.serviceList = serviceEnabledList
+	componentEnabledObj.namespaceList = namespaceEnabledList
+	return &componentEnabledObj, nil
+}
+
+func (c *ComponentEnabled) containsService(service string, ns string) bool {
+	for _, item := range c.serviceList {
```
mcieplak, line 473: Should we make this a map with the key being namespace/service so the lookup is O(1)?

pkg/istio/rbac/common/util_test.go
```diff
@@ -1018,3 +1018,185 @@ func getAuthzPolicyItem(action model.Event) Item {
...
+		components, err := ParseComponentsEnabledAuthzPolicy(testcase.input.objectList)
+		if err != nil {
+			if err.Error() != testcase.output.err {
+				t.Errorf("Wrong error message. Expected: %s, Actual: %s", testcase.output.err, err.Error())
+			} else {
+				continue
+			}
+		}
+
+		if len(components.serviceList) != len(testcase.output.result.serviceList) {
```
mcieplak, line 1105: I believe we can simplify all these checks using assert.EqualValues to do the comparisons here.

mcieplak (changes requested):

pkg/istio/rbac/common/util.go
```diff
@@ -421,3 +421,77 @@ func FetchServicesFromDir(namespace, localDirPath string) ([]string, error) {
...
+
+// ServiceEnabled - service and namespace combination that enabled authz policy
+type ServiceEnabled struct {
+	service   string
+	namespace string
+}
+
+type ComponentEnabled struct {
+	serviceMap    map[ServiceEnabled]string
+	namespaceMap  map[string]string
```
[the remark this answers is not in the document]
mcieplak: +1 I think we should just make it a bool value.

pkg/istio/rbac/common/util.go
```diff
@@ -429,8 +429,8 @@ type ServiceEnabled struct {
 }
 
 type ComponentEnabled struct {
-	serviceList   []ServiceEnabled
-	namespaceList []string
+	serviceMap    map[ServiceEnabled]string
```
mcieplak, line 432: I think it will be easier to just make this `map[string]bool` instead of `map[ServiceEnabled]string`. That way the key will just be `namespace/service` instead of the entire struct object. It will also simplify the lookup as well since we don't need to create an object for lookup.

main.go
```diff
@@ -34,7 +34,7 @@ func main() {
 	logFile := flag.String("log-file", "/var/log/k8s-athenz-istio-auth/k8s-athenz-istio-auth.log", "log file location")
 	logLevel := flag.String("log-level", "info", "logging level")
 	enableAuthzPolicyController := flag.Bool("enable-authzpolicy-controller", true, "enable authzpolicy controller to create authzpolicy dry run resource")
```
mcieplak, line 36: Should we make this arg `enable-ap-controller` to be consistend with `ap-enabled-list`?

pkg/istio/rbac/v2/provider_test.go
```diff
@@ -55,7 +56,19 @@ func TestConvertAthenzModelIntoIstioRbac(t *testing.T) {
 
 	convertedCR := p.ConvertAthenzModelIntoIstioRbac(domainRBAC, onboardedService.Name, labels["app"])
 	expectedCR := getExpectedCR()
-	assert.Equal(t, expectedCR, convertedCR, "converted authz policy should be equal")
+	// when there are mulitple assertions matched with svc,
```
mcieplak, line 59: Typo: multiple.

pkg/istio/rbac/v2/provider_test.go
```diff
@@ -55,7 +56,19 @@ func TestConvertAthenzModelIntoIstioRbac(t *testing.T) {
...
+	// base on which assertion will be processed from the athenzModel mapping, order of the authz rule generated
+	// may be different each time. Using a workaround here to sort the configSpec in alphabetical order of first
+	// method in each rule.
+	configSpec := (convertedCR[0].Spec).(*v1beta1.AuthorizationPolicy)
+	rules := configSpec.Rules
+	sort.Slice(rules, func(i, j int) bool {
```
mcieplak, line 65: I believe sort.Slice directly modifies the underlying array so we could just call this function using configSpec.Rules and remove the temporary variables.

mcieplak (changes requested):

mcieplak (approved):

mdhume (approved):

PR #51, opened 2021-03-09
Add sync error check, add length check for request principals append

mcieplak:

@RuiGuo34 @mdhume

### DIFF
--- a/pkg/istio/authorizationpolicy/controller_test.go
+++ b/pkg/istio/authorizationpolicy/controller_test.go
@@ -59,7 +59,7 @@ var (
 		},
 	}
 
-	notOnboardedServiceWithAnnotationTrue = &v1.Service{
+	undefinedAthenzRulesServiceWithAnnotationTrue = &v1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "onboarded-service",
 			Namespace: "test-namespace-not-onboarded",
@@ -248,11 +248,11 @@ func TestSyncService(t *testing.T) {
 		},
 		{
 			name:                "create empty Authorization Policy spec when there is create event of service which doesn't have roles / policies defined",
-			inputService:        notOnboardedServiceWithAnnotationTrue,
+			inputService:        undefinedAthenzRulesServiceWithAnnotationTrue,
 			inputAthenzDomain:   notOnboardedAthenzDomain,
 			existingAuthzPolicy: nil,
 			expectedAuthzPolicy: getExpectedEmptyAuthzPolicy(),
-			item:                Item{Operation: model.EventAdd, Resource: notOnboardedServiceWithAnnotationTrue},
+			item:                Item{Operation: model.EventAdd, Resource: undefinedAthenzRulesServiceWithAnnotationTrue},
 		},
 	}
 
--- a/pkg/istio/authorizationpolicy/controller_test.go
+++ b/pkg/istio/authorizationpolicy/controller_test.go
@@ -59,7 +59,20 @@ var (
 		},
 	}
 
-	notOnboardedServiceWithAnnotation = &v1.Service{
+	notOnboardedServiceWithAnnotationTrue = &v1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "onboarded-service",
+			Namespace: "test-namespace-not-onboarded",
+			Annotations: map[string]string{
+				authzEnabledAnnotation: "true",
+			},
+			Labels: map[string]string{
+				"svc": "productpage",
+			},
+		},
+	}
+
+	notOnboardedServiceWithAnnotationFalse = &v1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "onboarded-service",
 			Namespace: "test-namespace-onboarded",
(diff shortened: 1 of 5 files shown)


pkg/istio/authorizationpolicy/controller_test.go
```diff
@@ -59,7 +59,20 @@ var (
 		},
 	}
 
-	notOnboardedServiceWithAnnotation = &v1.Service{
+	notOnboardedServiceWithAnnotationTrue = &v1.Service{
```
mdhume, line 62: is this when roles/policies are not defined for the service but the authz flag in omega.yaml is set to true? If so, can we rename the test to something like `undefinedAthenzRulesServiceWithAnnotationTrue`

pkg/istio/rbac/v2/provider_test.go
```diff
@@ -40,23 +40,76 @@ var (
 			},
 		},
 	}
+
+	notOnboardedServiceWithAnnotationTrue = &k8sv1.Service{
```
mdhume, line 44: same comment as above

pkg/istio/authorizationpolicy/controller_test.go
mcieplak: Yep, updated.

pkg/istio/rbac/v2/provider_test.go
mcieplak: Updated.

mdhume (approved):

pkg/istio/rbac/v2/provider_test.go
```diff
@@ -40,23 +40,76 @@ var (
...
+	}
+}
 
-	convertedCR := p.ConvertAthenzModelIntoIstioRbac(domainRBAC, onboardedService.Name, labels["app"])
-	expectedCR := getExpectedCR()
-	assert.Equal(t, expectedCR, convertedCR, "converted authz policy should be equal")
+func getExpectedEmptyAuthzPolicy() []model.Config {
+	var out model.Config
+	schema := collections.IstioSecurityV1Beta1Authorizationpolicies
+	createTimestamp, _ := time.Parse("", "12/8/2015 12:00:00")
```
RuiGuo34, line 95: let's do an error as we did elsewhere.

pkg/istio/authorizationpolicy/controller_test.go
```diff
@@ -227,12 +240,20 @@ func TestSyncService(t *testing.T) {
...
 			name:                "create Authorization Policy spec when there is update event of service from no annotation set to annotation set",
-			inputService:        notOnboardedServiceWithAnnotation,
+			inputService:        notOnboardedServiceWithAnnotationFalse,
 			inputAthenzDomain:   onboardedAthenzDomain,
 			existingAuthzPolicy: nil,
 			expectedAuthzPolicy: getExpectedAuthzPolicy(),
 			item:                Item{Operation: model.EventUpdate, Resource: onboardedService},
 		},
+		{
+			name:                "create empty Authorization Policy spec when there is create event of service which doesn't have roles / policies defined",
```
RuiGuo34, line 250: condition is: doesn't have both roles and policies defined.
When there are roles but no policies, or when there are policies but no roles. Will the conversion behave correctly? Do you think we can add these two conditions as test cases as well?

pkg/istio/rbac/v2/provider_test.go
```diff
@@ -40,23 +40,76 @@ var (
...
+		},
+	}
+	return []model.Config{out}
 }
 
-func getExpectedCR() []model.Config {
+func getExpectedAuthzPolicy() []model.Config {
 	var out model.Config
 	schema := collections.IstioSecurityV1Beta1Authorizationpolicies
 	createTimestamp, _ := time.Parse("", "12/8/2015 12:00:00")
```
RuiGuo34, line 115: leave as a placeholder, err should be checked from upstream branch, double check after pull from branch authzpolicy

pkg/istio/rbac/v2/provider_test.go
mcieplak: Added.

pkg/istio/authorizationpolicy/controller_test.go
mcieplak: Sure, added test cases to cover all the different scenarios.

pkg/istio/rbac/v2/provider_test.go
mcieplak: Added.

pkg/istio/rbac/v2/provider_test.go
```diff
@@ -42,37 +42,159 @@ var (
...
+			name:                "should create empty authz policy spec for service with empty role and empty policy defined",
+			inputAthenzDomain:   getFakeNotOnboardedDomain(true, false, true, false),
+			inputService:        undefinedAthenzRulesServiceWithAnnotationTrue,
+			expectedAuthzPolicy: getExpectedEmptyAuthzPolicy(),
+		},
+		{
+			name:                "should create authz policy spec with role for service with no role and policy defined",
+			inputAthenzDomain:   getFakeNotOnboardedDomain(false, false, true, true),
+			inputService:        undefinedAthenzRulesServiceWithAnnotationTrue,
+			expectedAuthzPolicy: getExpectedEmptyAuthzPolicyWithRole(),
```
mdhume, line 101: minor: could we call this `getExpectedAuthzPolicyWithRole` since it is no longer an empty authz policy that is being returned

RuiGuo34 (approved):

pkg/istio/rbac/v2/provider_test.go
mcieplak: Sure, updated.

mdhume (approved):

Issue #56, opened 2021-03-20
Support for K8s namespace level authorization in AuthorizationPolicy


mdhume:
**Please describe the problem that you are trying to solve**
[Istio Authorization policies](https://istio.io/latest/docs/reference/config/security/authorization-policy/) have support for authorizing all the services from a given namespace using an additional field [v1beta1.Rule_From.Source.Namespaces](https://istio.io/latest/docs/reference/config/security/authorization-policy/#Source). If users provide an [Athenz role member](https://athenz.github.io/athenz/data_model/#roles) in the form of `<athenz-domain>.*` we should add the field `v1beta1.Rule_From.Source.Namespaces` to the corresponding AuthorizationPolicy to enable it.

**Describe the solution you'd like**

1. Update the [ConvertAthenzModelIntoIstioRbac](https://github.com/yahoo/k8s-athenz-istio-auth/blob/ed59085629060a3225391780e771a0fcfc4ecc47/pkg/istio/rbac/v2/provider.go#L34) to check the value of the role member. 
2. If the value of the role member:
    * contains a wildcard character `*` in the suffix
    * does not contain `user.` as the prefix
    then skip appending to `v1beta1.Rule_From.Source.Principals`
3. Convert the Athenz domain name into the corresponding Kubernetes namespace using the [DomainToNamespace](https://github.com/yahoo/k8s-athenz-istio-auth/blob/99afa06eb0882a35df486bb864066c313d47b366/pkg/athenz/util.go#L11) function
4. Create a `v1beta1.Rule_From.Source.Namespaces` array and append the generated namespace value to it

patrasap0908:
Hi @mdhume, I'd like to take this one if that's okay.

mcieplak:
PR has been merged, closing this issue. Thanks @patrasap0908 for your contribution!!

PR #58, opened 2021-03-21
Add support for namespace-level authorization in AuthorizationPolicy

patrasap0908:
This PR adds support for authorizing entire K8s namespaces when role members on Athenz are of the form `<athenz-domain>.*`, as described by https://github.com/yahoo/k8s-athenz-istio-auth/issues/56.

Expected translation:
**Role member** like `test-domain.namespace.*` leads to  the **K8s namespace** named `test--domain-namespace` getting added to the list of sources in the authz policy's rules.

### DIFF
--- a/pkg/istio/rbac/common/util.go
+++ b/pkg/istio/rbac/common/util.go
@@ -5,20 +5,22 @@ package common
 
 import (
 	"fmt"
-	"github.com/ghodss/yaml"
-	"github.com/gogo/protobuf/proto"
-	"github.com/yahoo/athenz/clients/go/zms"
-	"github.com/yahoo/k8s-athenz-istio-auth/pkg/log"
 	"io/ioutil"
-	"istio.io/istio/pilot/pkg/config/kube/crd"
-	"istio.io/istio/pilot/pkg/model"
-	"istio.io/istio/pkg/config/schema/collection"
-	"istio.io/istio/pkg/config/schema/collections"
 	"net/http"
 	"os"
 	"regexp"
 	"strings"
 	"time"
+
+	"github.com/ghodss/yaml"
+	"github.com/gogo/protobuf/proto"
+	"github.com/yahoo/athenz/clients/go/zms"
+	"github.com/yahoo/k8s-athenz-istio-auth/pkg/athenz"
+	"github.com/yahoo/k8s-athenz-istio-auth/pkg/log"
+	"istio.io/istio/pilot/pkg/config/kube/crd"
+	"istio.io/istio/pilot/pkg/model"
+	"istio.io/istio/pkg/config/schema/collection"
+	"istio.io/istio/pkg/config/schema/collections"
 )
 
 const (
@@ -94,6 +96,24 @@ func (a *ApiHandler) Delete(item *Item) error {
 	res := item.Resource
 	err := a.ConfigStoreCache.Delete(collections.IstioSecurityV1Beta1Authorizationpolicies.Resource().GroupVersionKind(), res.Name, res.Namespace)
 	return err
+}
+
+// CheckIfMemberIsAllUsers returns namespace for Athenz domain when role member is of form '<athenz-domain>.*'.
+// Example: domain.sub-domain.* -> domain-sub--domain
+func CheckIfMemberIsAllUsers(member *zms.RoleMember, domainName zms.DomainName) (string, error) {
+
+	if member == nil {
+		return "", fmt.Errorf("member is nil")
+	}
+
+	memberStr := string(member.MemberName)
+
+	// if member name is of the form '<athenz-domain>.*', return namespace
+	if strings.HasPrefix(memberStr, "user.") || !strings.HasSuffix(memberStr, ".*") {
+		return "", nil
+	}
+
+	return athenz.DomainToNamespace(memberStr[0 : len(memberStr)-2]), nil
 }
 
 // MemberToSpiffe parses the Athenz role member into a SPIFFE compliant name.
--- a/pkg/istio/rbac/common/util.go
+++ b/pkg/istio/rbac/common/util.go
@@ -109,7 +109,7 @@ func CheckIfMemberIsAllUsers(member *zms.RoleMember, domainName zms.DomainName) 
 	memberStr := string(member.MemberName)
 
 	// if member name is of the form '<athenz-domain>.*', return namespace
-	if strings.HasPrefix(memberStr, "user.") || !strings.HasSuffix(memberStr, ".*") {
+	if strings.HasPrefix(memberStr, "unix.") || strings.HasPrefix(memberStr, "user.") || !strings.HasSuffix(memberStr, ".*") {
 		return "", nil
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/istio/rbac/common/util.go
```diff
@@ -96,6 +98,24 @@ func (a *ApiHandler) Delete(item *Item) error {
...
 
+// CheckIfMemberIsAllUsers returns namespace for Athenz domain when role member is of form '<athenz-domain>.*'.
+// Example: domain.sub-domain.* -> domain-sub--domain
+func CheckIfMemberIsAllUsers(member *zms.RoleMember, domainName zms.DomainName) (string, error) {
+
+	if member == nil {
+		return "", fmt.Errorf("member is nil")
+	}
+
+	memberStr := string(member.MemberName)
```
RuiGuo34, line 109: there is no need to do string cast since MemberName is already type String

pkg/istio/rbac/common/util.go
```diff
@@ -96,6 +98,24 @@ func (a *ApiHandler) Delete(item *Item) error {
...
+
+	// if member name is of the form '<athenz-domain>.*', return namespace
+	if strings.HasPrefix(memberStr, "user.") || !strings.HasSuffix(memberStr, ".*") {
```
RuiGuo34, line 112: I think there are some cases like `cd.screwdriver.project.*` or `k8s.omega.*` could not be captured here. Should we do a strict check, equal to * here?

pkg/istio/rbac/common/util.go
```diff
@@ -96,6 +98,24 @@ func (a *ApiHandler) Delete(item *Item) error {
...
+		return "", nil
+	}
+
+	return athenz.DomainToNamespace(memberStr[0 : len(memberStr)-2]), nil
```
RuiGuo34, line 116: you can avoid string parsing by just calling athenz.DomainToNamespace(domainName)

pkg/istio/rbac/common/util.go
patrasap0908: Why wouldn't they get handled here? Both of those cases don't have `user.` as a prefix and have `.*` as the suffix.

pkg/istio/rbac/common/util.go
patrasap0908: `MemberName` is of type `zms.MemberName` and needs to be converted to a string before proceeding.

pkg/istio/rbac/common/util.go
patrasap0908: We can't use `domainName` here because that's the domain of the role whose members are currently getting processed, and all members won't necessarily belong to the same domain.

pkg/istio/rbac/v2/provider_test.go
```diff
@@ -218,10 +220,18 @@ func getExpectedAuthzPolicy() []model.Config {
...
+							Namespaces: []string{
+								"test--domain-namespace2",
+							},
+						},
+					},
 					{
 						Source: &v1beta1.Source{
 							RequestPrincipals: []string{
 								"*",
+								"athenz/test-domain.namespace2.*",
```
RuiGuo34, line 234: role for the namespace should not be added as `RequestPrincipals`, can you update code for generating `RequestPrincipals` to exclude namespace role?

pkg/istio/rbac/common/util.go
RuiGuo34: What I am trying to say is if there is non real user regex `cd.screwdriver.project.*`, the logic here will convert it to a namespace rule.

pkg/istio/rbac/common/util.go
RuiGuo34: Makes sense

pkg/istio/rbac/common/util.go
RuiGuo34: Check with @mcieplak for athenz user naming convention, there won't be above case.

RuiGuo34: Can you also check for prefix `unix.`? a valid user name in athenz can be `unix.<group>`,  we want to avoid creating namespace in this case.

patrasap0908: Isn't `cd-screwdriver-project` a K8s namespace? A namespace for all Screwdriver services?

patrasap0908: Not sure what those SD services would get mapped to in the K8s space though.

pkg/istio/rbac/v2/provider_test.go
patrasap0908: https://github.com/yahoo/k8s-athenz-istio-auth/pull/58/commits/c2131ccc31aa4edd6d200d83c5067f928a93cb2c

pkg/istio/rbac/common/util.go
patrasap0908: For bypassing `unix.*` - https://github.com/yahoo/k8s-athenz-istio-auth/pull/58/commits/c2131ccc31aa4edd6d200d83c5067f928a93cb2c.

RuiGuo34 (approved):
LGTM

pkg/istio/rbac/common/util.go
```diff
@@ -96,6 +98,24 @@ func (a *ApiHandler) Delete(item *Item) error {
 	return err
 }
 
+// CheckIfMemberIsAllUsers returns namespace for Athenz domain when role member is of form '<athenz-domain>.*'.
+// Example: domain.sub-domain.* -> domain-sub--domain
+func CheckIfMemberIsAllUsers(member *zms.RoleMember, domainName zms.DomainName) (string, error) {
```
mcieplak, line 103: Could we rename this to `CheckIfMemberIsAllUsersFromDomain` or something along those lines? I think AllUsers is a little bit confusing because we also support `user.*` which allows all athenz users and services access.

pkg/istio/rbac/v2/provider_test.go
```diff
@@ -18,15 +22,13 @@ import (
...
-	"sort"
-	"testing"
-	"time"
 )
 
 const (
 	domainName       = "test.namespace"
 	username         = "user.name"
 	wildcardUsername = "user.*"
+	wildcardAllUsers = "test-domain.namespace2.*"
```
mcieplak, line 31: Could we also rename this to `wildcardAllDomainUsers` or something along those lines?

pkg/istio/rbac/common/util.go
patrasap0908: Agreed, done - https://github.com/yahoo/k8s-athenz-istio-auth/pull/58/commits/db86ddc3af039aebaa3d8d72ec673964855b4ce2.

pkg/istio/rbac/v2/provider_test.go
patrasap0908: https://github.com/yahoo/k8s-athenz-istio-auth/pull/58/commits/db86ddc3af039aebaa3d8d72ec673964855b4ce2

mcieplak (approved):

RuiGuo34 (approved):

PR #59, opened 2021-03-26
Adding logic to drop query parameters when using Authz Policy

kpharasi:

@mcieplak @RuiGuo34

### DIFF
--- a/pkg/istio/rbac/v2/provider.go
+++ b/pkg/istio/rbac/v2/provider.go
@@ -4,9 +4,11 @@
 package v2
 
 import (
-	"github.com/yahoo/athenz/clients/go/zms"
 	"regexp"
 	"sort"
+	"strings"
+
+	"github.com/yahoo/athenz/clients/go/zms"
 
 	"github.com/yahoo/k8s-athenz-istio-auth/pkg/athenz"
 	"github.com/yahoo/k8s-athenz-istio-auth/pkg/istio/rbac"
@@ -78,6 +80,23 @@ func (p *v2) ConvertAthenzModelIntoIstioRbac(athenzModel athenz.Model, serviceNa
 			if err != nil {
 				continue
 			}
+
+			// Drop the query parameters from the HTTP path in the assertions due to the difference
+			// in the RBAC Envoy permissions config created by Authentication Policy and ServiceRole/ServiceRoleBindings.
+			// Which in case of,
+			// Authentication Policy - is created with a url_path object
+			// ServiceRole/ServiceRoleBindings - is created with a header object
+			re, err := regexp.Compile(`.*\?.*`)
+			if err != nil {
+				log.Errorln("Error while parsing the path regex: ", err.Error())
+				continue
+			}
+
+			if re.MatchString(path) {
+				pathArr := strings.Split(path, "?")
+				path = pathArr[0]
+			}
+
 			// if svc match with current svc, process it and add it to the rules
 			// note that svc defined on athenz can be a regex, need to match the pattern
 			res, err := regexp.MatchString(svc, svcLabel)
--- a/pkg/istio/rbac/v2/provider.go
+++ b/pkg/istio/rbac/v2/provider.go
@@ -33,6 +33,9 @@ func NewProvider(componentEnabledAuthzPolicy *common.ComponentEnabled, enableOri
 	}
 }
 
+// Regex for finding if the HTTP path contains a query parameter
+var re = regexp.MustCompile(`.*\?.*`)
+
 // ConvertAthenzModelIntoIstioRbac converts the Athenz RBAC model into Istio Authorization V1Beta1 specific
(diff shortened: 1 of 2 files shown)


pkg/istio/rbac/v2/provider.go
```diff
@@ -78,6 +80,23 @@ func (p *v2) ConvertAthenzModelIntoIstioRbac(athenzModel athenz.Model, serviceNa
 			if err != nil {
 				continue
 			}
+
+			// Drop the query parameters from the HTTP path in the assertions due to the difference
+			// in the RBAC Envoy permissions config created by Authentication Policy and ServiceRole/ServiceRoleBindings.
+			// Which in case of,
+			// Authentication Policy - is created with a url_path object
+			// ServiceRole/ServiceRoleBindings - is created with a header object
+			re, err := regexp.Compile(`.*\?.*`)
```
RuiGuo34, line 89: Can you move this block after checking svc regex? This will reduce # of operations.

pkg/istio/rbac/v2/provider.go
```diff
@@ -78,6 +80,23 @@ func (p *v2) ConvertAthenzModelIntoIstioRbac(athenzModel athenz.Model, serviceNa
...
+			if err != nil {
+				log.Errorln("Error while parsing the path regex: ", err.Error())
+				continue
+			}
+
+			if re.MatchString(path) {
```
RuiGuo34, line 95: Can you add a simple test to cover path with query param?

kpharasi: Sure, will do

pkg/istio/rbac/v2/provider.go
kpharasi: Sure, makes sense

mcieplak: Since this regex is the same for every iteration, can we move the compilation out of this function entirely so that it only needs to happen once? Maybe as a variable at the top of this file?

pkg/istio/rbac/v2/provider.go
mcieplak: +1

pkg/istio/rbac/v2/provider_test.go
```diff
@@ -310,6 +321,12 @@ func getFakeOnboardedDomain() zms.SignedDomain {
 									Action:   "post",
 									Effect:   &allow,
 								},
+								{
+									Role:     domainName + ":role.productpage-writer",
+									Resource: domainName + ":svc.productpage:/api/query?*",
+									Action:   "post",
+									Effect:   &allow,
+								},
```
RuiGuo34, line 329: could you add one more test case like `?name=test`, case which substring after `?` is not a regex?

RuiGuo34 (approved):

pkg/istio/rbac/v2/provider.go
```diff
@@ -78,6 +83,17 @@ func (p *v2) ConvertAthenzModelIntoIstioRbac(athenzModel athenz.Model, serviceNa
 			if err != nil {
 				continue
 			}
+
+			// Drop the query parameters from the HTTP path in the assertions due to the difference
+			// in the RBAC Envoy permissions config created by Authentication Policy and ServiceRole/ServiceRoleBindings.
```
mcieplak, line 88: This should be authorization policy as authentication policy is a different resource used for mTLS etc.

pkg/istio/rbac/v2/provider.go
```diff
@@ -31,6 +33,9 @@ func NewProvider(componentEnabledAuthzPolicy *common.ComponentEnabled, enableOri
 	}
 }
 
+// Regex for finding if the HTTP path contains a query parameter
+var re = regexp.MustCompile(`.*\?.*`)
```
mcieplak, line 37: Could we rename this to queryRegex?

mcieplak (changes requested):

pkg/istio/rbac/v2/provider.go
```diff
@@ -78,6 +83,17 @@ func (p *v2) ConvertAthenzModelIntoIstioRbac(athenzModel athenz.Model, serviceNa
 			if err != nil {
 				continue
 			}
+
+			// Drop the query parameters from the HTTP path in the assertions due to the difference
+			// in the RBAC Envoy permissions config created by Authentication Policy and ServiceRole/ServiceRoleBindings.
+			// Which in case of,
+			// Authentication Policy - is created with a url_path object
```
mcieplak, line 90: Can we update to authorization policy here as well?

pkg/istio/rbac/v2/provider.go
kpharasi: Ah, ya. Thanks for catching that

mcieplak (approved):

PR #62, opened 2021-04-01
Handle services which are flipping from AP->SR/SRB and SR/SRB->AP

kpharasi:

@mcieplak @mdhume @RuiGuo34

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,6 +5,12 @@ package main
 
 import (
 	"flag"
+	"os"
+	"os/signal"
+	"path/filepath"
+	"syscall"
+	"time"
+
 	"github.com/yahoo/k8s-athenz-istio-auth/pkg/controller"
 	"github.com/yahoo/k8s-athenz-istio-auth/pkg/istio/rbac/common"
 	"github.com/yahoo/k8s-athenz-istio-auth/pkg/log"
@@ -17,11 +23,6 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/util/homedir"
-	"os"
-	"os/signal"
-	"path/filepath"
-	"syscall"
-	"time"
 )
 
 func main() {
@@ -39,12 +40,21 @@ func main() {
 	flag.Parse()
 	log.InitLogger(*logFile, *logLevel)
 
+	// When enableAuthzPolicyController is set to true create a dry run folder which
+	// would contain the Authorization Policy resource for all the namespaces/services which
+	// are not passed as a parameter in --ap-enabled-list
 	if *enableAuthzPolicyController {
-		if _, err := os.Stat(common.DryRunStoredFilesDirectory); os.IsNotExist(err) {
-			err := os.MkdirAll(common.DryRunStoredFilesDirectory, 0755)
+		// If the Authz Policy folder already exists remove the stale data
+		if _, err := os.Stat(common.DryRunStoredFilesDirectory); !os.IsNotExist(err) {
+			err := os.RemoveAll(common.DryRunStoredFilesDirectory)
 			if err != nil {
-				log.Panicf("Error when creating authz policy directory: %s", err.Error())
+				log.Panicf("Error when removing authz policy directory: %s", err.Error())
 			}
+		}
+
+		err := os.MkdirAll(common.DryRunStoredFilesDirectory, 0755)
+		if err != nil {
+			log.Panicf("Error when creating authz policy directory: %s", err.Error())
 		}
 	}
 
@@ -57,7 +67,7 @@ func main() {
 		}
 	}
 
-	//Ledger for tracking config distribution, specify how long it can retain its previous state
+	// Ledger for tracking config distribution, specify how long it can retain its previous state
 	configLedger := ledger.Make(time.Hour)
 	istioClient, err := crdController.NewClient(*kubeconfig, "", configDescriptor, *dnsSuffix, configLedger, "")
 	if err != nil {
@@ -96,6 +106,8 @@ func main() {
 		log.Panicf("Error parsing ap-resync-interval duration: %s", err.Error())
 	}
 
+	// When enableAuthzPolicyController is set to true determine which services,
+	// namespaces or cluster to create Authorization Policies for
 	var componentsEnabledAuthzPolicy *common.ComponentEnabled
 	if *enableAuthzPolicyController {
 		componentsEnabledAuthzPolicy, err = common.ParseComponentsEnabledAuthzPolicy(*authzPolicyEnabledList)
--- a/main.go
+++ b/main.go
@@ -45,7 +45,7 @@ func main() {
 	// are not passed as a parameter in --ap-enabled-list
 	if *enableAuthzPolicyController {
 		// If the Authz Policy folder already exists remove the stale data
-		if _, err := os.Stat(common.DryRunStoredFilesDirectory); !os.IsNotExist(err) {
+		if _, err := os.Stat(common.DryRunStoredFilesDirectory); err == nil {
 			err := os.RemoveAll(common.DryRunStoredFilesDirectory)
 			if err != nil {
 				log.Panicf("Error when removing authz policy directory: %s", err.Error())
--- a/main.go
+++ b/main.go
@@ -50,6 +50,8 @@ func main() {
 			if err != nil {
 				log.Panicf("Error when removing authz policy directory: %s", err.Error())
 			}
+		} else if !os.IsNotExist(err) {
+			log.Panicf("Error when checking for the presence of the dry run directory: %s", err.Error())
 		}
 
 		err := os.MkdirAll(common.DryRunStoredFilesDirectory, 0755)
--- a/pkg/istio/authorizationpolicy/controller.go
+++ b/pkg/istio/authorizationpolicy/controller.go
@@ -118,13 +118,15 @@ func (c *Controller) Run(stopCh <-chan struct{}) {
 	go c.authzpolicyIndexInformer.Run(stopCh)
 	go c.resync(stopCh)
 
-	if c.configStoreCache.HasSynced() {
-		// If the service is switching back from Authorization Policy Enabled back to SR/SRB delete the
-		// existing Authorization Policy associated to the service
-		err := c.cleanUpStaleAP()
-		if err != nil {
-			log.Errorf("Error while running cleanUpStaleAP: %v", err.Error())
-		}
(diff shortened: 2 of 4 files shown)


main.go
```diff
@@ -39,13 +40,22 @@ func main() {
...
 	log.InitLogger(*logFile, *logLevel)
 
+	// When enableAuthzPolicyController is set to true create a dry run folder which
+	// would contain the Authorization Policy resource for all the namespaces/services which
+	// are not passed as a parameter in --ap-enabled-list
 	if *enableAuthzPolicyController {
-		if _, err := os.Stat(common.DryRunStoredFilesDirectory); os.IsNotExist(err) {
-			err := os.MkdirAll(common.DryRunStoredFilesDirectory, 0755)
+		// If the Authz Policy folder already exists remove the stale data
+		if _, err := os.Stat(common.DryRunStoredFilesDirectory); !os.IsNotExist(err) {
```
RuiGuo34, line 48: can you switch to os.IsExist()?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -117,6 +118,15 @@ func (c *Controller) Run(stopCh <-chan struct{}) {
 	go c.authzpolicyIndexInformer.Run(stopCh)
 	go c.resync(stopCh)
 
+	if c.configStoreCache.HasSynced() {
+		// If the service is switching back from Authorization Policy Enabled back to SR/SRB delete the
+		// existing Authorization Policy associated to the service
+		err := c.cleanUpStaleAP()
+		if err != nil {
+			log.Errorf("Error while running cleanUpStaleAP: %v", err.Error())
```
RuiGuo34, line 126: If clean up failed with problem, do we want to panic exiting the controller?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -369,3 +385,43 @@ func (c *Controller) checkAuthzEnabledAnnotation(serviceObj *corev1.Service) boo
...
+		nsSvcArr := strings.Split(currAP.Key(), "/")
+		serviceName := nsSvcArr[2]
+		serviceNamespace := nsSvcArr[1]
+
+		// Creating the Item to pass to the apiHandler
+		// with a delete event
+		key := serviceNamespace + "/" + serviceName
+		cbHandler := c.getCallbackHandler(key)
+
+		item := &common.Item{
```
RuiGuo34, line 408: item can be formed after checking if current ap is disabled

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -369,3 +385,43 @@ func (c *Controller) checkAuthzEnabledAnnotation(serviceObj *corev1.Service) boo
...
+			Resource:        currAP,
+			CallbackHandler: cbHandler,
+		}
+
+		eHandler := &c.apiHandler
+
+		// Check if the Authorization Policy is enabled for the service through ap-enabled-list
+		if !c.componentEnabledAuthzPolicy.IsEnabled(serviceName, serviceNamespace) {
+			log.Infof("Deleting stale AP, namespace: %v service name: %v", serviceNamespace, serviceName)
+			err = eHandler.Delete(item)
```
RuiGuo34, line 419: can be simplified by calling c.apiHandler.Delete()

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -369,3 +385,43 @@ func (c *Controller) checkAuthzEnabledAnnotation(serviceObj *corev1.Service) boo
...
+			Operation:       model.EventDelete,
+			Resource:        currAP,
+			CallbackHandler: cbHandler,
+		}
+
+		eHandler := &c.apiHandler
+
+		// Check if the Authorization Policy is enabled for the service through ap-enabled-list
+		if !c.componentEnabledAuthzPolicy.IsEnabled(serviceName, serviceNamespace) {
+			log.Infof("Deleting stale AP, namespace: %v service name: %v", serviceNamespace, serviceName)
```
RuiGuo34, line 418: Before deleting, need to check if ap has override annotation set, we don't want to delete user's manually applied ap.

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -369,3 +385,43 @@ func (c *Controller) checkAuthzEnabledAnnotation(serviceObj *corev1.Service) boo
 	}
 	return false
 }
+
+// cleanUpStaleAP deletes the existing Authorization Policy associated to the service which is switching back from
+// Authorization Policy Enabled back to SR/SRB
+func (c *Controller) cleanUpStaleAP() error {
```
RuiGuo34, line 391: Can we add a unit test here? If not, can it be added as an integration test?

main.go
kpharasi: `os.IsExist(err)` would not work here. `os.IsNotExist(err)` & `os.IsExist(err)` both rely on the err message from `os.Stat` to determine if the file exists or not. In case the file exists `os.Stat` will return nil and `os.IsExist` would thus return false (as err is nil). Thus, we make use of `!os.IsNotExist(err)`.

kpharasi: Actually, let me change that to `err==nil`

pkg/istio/authorizationpolicy/controller.go
kpharasi: Will add a unit test

pkg/istio/authorizationpolicy/controller.go
kpharasi: Currently, it does not but I can change it to error out in case the cleanup failed. Can not think of any cons with that. Thoughts @mdhume @mcieplak ?

mcieplak: I think it makes sense to panic, this will at least alert us that something is wrong. Otherwise if we move on then we might run into a weird state with some authz policies still on the clusters that have now moved back to sr / srb.

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -117,6 +118,15 @@ func (c *Controller) Run(stopCh <-chan struct{}) {
 	go c.authzpolicyIndexInformer.Run(stopCh)
 	go c.resync(stopCh)
 
+	if c.configStoreCache.HasSynced() {
```
mcieplak, line 121: Can we actually add a wait for cache sync here instead of an if statement like we do here: https://github.com/yahoo/k8s-athenz-istio-auth/blob/master/pkg/controller/controller.go#L216-L218
We need to wait until it has fully synced before processing anything. This if statement will just check it once.

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -369,3 +387,41 @@ func (c *Controller) checkAuthzEnabledAnnotation(serviceObj *corev1.Service) boo
...
+	currentAPList, err := c.configStoreCache.List(collections.IstioSecurityV1Beta1Authorizationpolicies.Resource().GroupVersionKind(), "")
+	if err != nil {
+		return fmt.Errorf("Error while fetching the Authorization Policy resources from cache : %v", err.Error())
+	}
+
+	for _, currAP := range currentAPList {
+		nsSvcArr := strings.Split(currAP.Key(), "/")
+		serviceName := nsSvcArr[2]
+		serviceNamespace := nsSvcArr[1]
+		key := serviceNamespace + "/" + serviceName
```
mcieplak, line 404: Can we just use `currApp.Namespace + "/" + currApp.Name` instead so we don't have to do any string manipulation?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -117,6 +118,17 @@ func (c *Controller) Run(stopCh <-chan struct{}) {
 	go c.authzpolicyIndexInformer.Run(stopCh)
 	go c.resync(stopCh)
 
+	if !cache.WaitForCacheSync(stopCh, c.configStoreCache.HasSynced, c.serviceIndexInformer.HasSynced, c.adIndexInformer.HasSynced) {
```
RuiGuo34, line 121: missed `c.authzpolicyIndexInformer`

pkg/istio/authorizationpolicy/controller_test.go
```diff
@@ -23,14 +27,11 @@ import (
 	"istio.io/istio/pkg/config/schema/collection"
 	"istio.io/istio/pkg/config/schema/collections"
 	"istio.io/istio/pkg/config/schema/resource"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
RuiGuo34, line 30: no need to explicitly specify v1

pkg/istio/authorizationpolicy/controller_test.go
```diff
@@ -497,6 +496,69 @@ func TestNewController(t *testing.T) {
...
+		{
+			name:                "existing authorization policy is not deleted as the same namespace as the existing service is part of apEnabledList",
+			inputService:        onboardedService,
+			inputAthenzDomain:   onboardedAthenzDomain,
+			existingAuthzPolicy: getExpectedAuthzPolicy(),
+			expectedAuthzPolicy: getExpectedAuthzPolicy(),
+			apEnabledList:       "test-namespace-onboarded/*",
+		},
+		{
+			name:                "existing authorization policy is not deleted as the override annotation is enabled even when not the same namespace",
```
mcieplak, line 533: I think we're missing a word here, `when not **in** the same namespace`

main.go
```diff
@@ -39,13 +40,22 @@ func main() {
...
 	log.InitLogger(*logFile, *logLevel)
 
+	// When enableAuthzPolicyController is set to true create a dry run folder which
+	// would contain the Authorization Policy resource for all the namespaces/services which
+	// are not passed as a parameter in --ap-enabled-list
 	if *enableAuthzPolicyController {
-		if _, err := os.Stat(common.DryRunStoredFilesDirectory); os.IsNotExist(err) {
-			err := os.MkdirAll(common.DryRunStoredFilesDirectory, 0755)
+		// If the Authz Policy folder already exists remove the stale data
+		if _, err := os.Stat(common.DryRunStoredFilesDirectory); err == nil {
```
mcieplak, line 48: Should we also handle the case where err is not nil?

pkg/istio/authorizationpolicy/controller.go
RuiGuo34: no need as we don't need authzpolicyIndexInformer

mcieplak (approved):

RuiGuo34 (approved):

PR #66, opened 2021-05-07
Adding support for Athenz Groups

kpharasi:
@mcieplak @RuiGuo34 @mdhume 

### DIFF
--- a/pkg/istio/rbac/common/util.go
+++ b/pkg/istio/rbac/common/util.go
@@ -289,7 +289,7 @@ func ParseAssertionResource(domainName zms.DomainName, assertion *zms.Assertion)
 // role member
 func CheckAthenzSystemDisabled(member interface{}) (bool, error) {
 	if member == nil {
-		return false, fmt.Errorf("got an empty role Member: %s, skipping", GetMemberName(member))
+		return false, fmt.Errorf("got an empty role Member, skipping")
 	}
 
 	systemDisabled := getMemberSystemDisabled(member)
@@ -304,7 +304,7 @@ func CheckAthenzSystemDisabled(member interface{}) (bool, error) {
 // checks if expiration date surpasses current time
 func CheckAthenzMemberExpiry(member interface{}) (bool, error) {
 	if member == nil {
-		return false, fmt.Errorf("got an empty role Member: %s, skipping", GetMemberName(member))
+		return false, fmt.Errorf("got an empty role Member, skipping")
 	}
 
 	expiration := getMemberExpiry(member)
--- a/pkg/istio/rbac/common/util.go
+++ b/pkg/istio/rbac/common/util.go
@@ -12,6 +12,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/ardielle/ardielle-go/rdl"
 	"github.com/ghodss/yaml"
 	"github.com/gogo/protobuf/proto"
 	"github.com/yahoo/athenz/clients/go/zms"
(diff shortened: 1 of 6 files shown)


pkg/istio/rbac/common/util.go
```diff
@@ -99,15 +100,57 @@ func (a *ApiHandler) Delete(item *Item) error {
...
+func getMemberExpiry(member interface{}) *rdl.Timestamp {
+	if groupMember, ok := member.(*zms.GroupMember); ok {
+		return groupMember.Expiration
+	}
+
+	if roleMember, ok := member.(*zms.RoleMember); ok {
+		return roleMember.Expiration
+	}
+
+	return &rdl.Timestamp{Time: time.Time{}}
```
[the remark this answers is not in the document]
RuiGuo34: can this simplify to &rdl.Timestamp{}, time.Time{} has a default value of `0001-01-01 00:00:00 +0000 UTC`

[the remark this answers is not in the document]
RuiGuo34: and when will we run into case to return default expiry? Could there be a case when a member is not of type *zms.GroupMember and *zms.RoleMember?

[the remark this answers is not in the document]
RuiGuo34: and when will we run into case to return a default expiry? could there be a case which a member is not of type *zms.GroupMember and *zms.RoleMember

pkg/istio/rbac/common/util.go
```diff
@@ -99,15 +100,57 @@ func (a *ApiHandler) Delete(item *Item) error {
...
+func getMemberSystemDisabled(member interface{}) *int32 {
+	if groupMember, ok := member.(*zms.GroupMember); ok {
+		return groupMember.SystemDisabled
+	}
 
+	if roleMember, ok := member.(*zms.RoleMember); ok {
+		return roleMember.SystemDisabled
+	}
+
+	return nil
```
RuiGuo34, line 143: have the same question as above

kpharasi: It should never be the case with our current setup. But this is in case we get something different and to handle that scenario.

pkg/istio/rbac/common/util.go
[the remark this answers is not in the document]
kpharasi: It should never be the case with our current setup. But this is in case we get something different and to handle that scenario. Will change the default to `nil`

pkg/istio/rbac/v2/provider.go
```diff
@@ -146,8 +146,27 @@ func (p *v2) ConvertAthenzModelIntoIstioRbac(athenzModel athenz.Model, serviceNa
 		from_namespace := &v1beta1.Rule_From{
 			Source: &v1beta1.Source{},
 		}
+
 		// role name should match zms resource name
 		for _, roleMember := range athenzModel.Members[role] {
+			var members []interface{}
+			roleflag := false
```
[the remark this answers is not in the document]
kpharasi: This is a flag I have used to differentiate if the type is *zms.GroupMember/*zms.RoleMember and to only run an extra check for expiry and system disabled in it is a *zms.GroupMember (Other than the *zms.RoleMember check)

pkg/istio/rbac/common/util.go
```diff
@@ -246,25 +287,31 @@ func ParseAssertionResource(domainName zms.DomainName, assertion *zms.Assertion)
 
 // CheckAthenzSystemDisabled checks if athenz domain is systematically disabled, if so, controller skips processing current
 // role member
-func CheckAthenzSystemDisabled(roleMember *zms.RoleMember) (bool, error) {
-	if roleMember == nil {
-		return false, fmt.Errorf("got an empty role Member: %s, skipping", roleMember.MemberName)
+func CheckAthenzSystemDisabled(member interface{}) (bool, error) {
+	if member == nil {
+		return false, fmt.Errorf("got an empty role Member: %s, skipping", GetMemberName(member))
```
mcieplak, line 292: Should we remove the `%s` from this string and also the GetMemberName call? Since it's nil, it will only return the empty string

pkg/istio/rbac/common/util.go
```diff
@@ -246,25 +287,31 @@ func ParseAssertionResource(domainName zms.DomainName, assertion *zms.Assertion)
...
 }
 
 // CheckAthenzMemberExpiry checks if Expiration field (timezone UTC) is set in the roleMember object, and then
 // checks if expiration date surpasses current time
-func CheckAthenzMemberExpiry(roleMember *zms.RoleMember) (bool, error) {
-	if roleMember == nil {
-		return false, fmt.Errorf("got an empty role Member: %s, skipping", roleMember.MemberName)
+func CheckAthenzMemberExpiry(member interface{}) (bool, error) {
+	if member == nil {
+		return false, fmt.Errorf("got an empty role Member: %s, skipping", GetMemberName(member))
```
mcieplak, line 307: Same here as well, won't this just be an empty string?

pkg/istio/rbac/common/util_test.go
```diff
@@ -166,6 +223,101 @@ func TestMemberToOriginJwtSubject(t *testing.T) {
...
+			},
+			expectedName: "client.some-domain.dep-svcA",
+		},
+		{
+			test: "valid group member",
+			member: &zms.GroupMember{
+				MemberName: zms.GroupMemberName("user.somename"),
+			},
+			expectedName: "user.somename",
+		},
```
mcieplak, line 246: Could we also add a case if member is nil?

pkg/istio/rbac/common/util_test.go
```diff
@@ -166,6 +223,101 @@ func TestMemberToOriginJwtSubject(t *testing.T) {
...
+			member: &zms.GroupMember{
+				MemberName: zms.GroupMemberName("user.somename"),
+				Expiration: &timestamp,
+			},
+			expectedExpiration: &timestamp,
+		},
+	}
+
+	for _, c := range cases {
+		expiry := getMemberExpiry(c.member)
```
mcieplak, line 285: Can we also add a test case if the member is nil?

pkg/istio/rbac/common/util_test.go
```diff
@@ -166,6 +223,101 @@ func TestMemberToOriginJwtSubject(t *testing.T) {
...
+			member: &zms.GroupMember{
+				MemberName:     zms.GroupMemberName("user.somename"),
+				SystemDisabled: &isNotSystemDisabled,
+			},
+			expectedSystemDisabled: &isNotSystemDisabled,
+		},
+	}
+
+	for _, c := range cases {
+		systemDisabled := getMemberSystemDisabled(c.member)
```
mcieplak, line 316: Can we also add the nil case here as well?

pkg/istio/rbac/v2/provider.go
```diff
@@ -166,28 +186,57 @@ func (p *v2) ConvertAthenzModelIntoIstioRbac(athenzModel athenz.Model, serviceNa
...
-				log.Errorln("error converting role member to spiffeName: ", err.Error())
-				continue
-			}
-			from_principal.Source.Principals = append(from_principal.Source.Principals, spiffeName)
-			if p.enableOriginJwtSubject {
-				originJwtName, err := common.MemberToOriginJwtSubject(roleMember)
+
+			for _, member := range members {
+				// This is only done for Group Members to check the expiry and system disabled
+				// at a Group Member level after doing a check for the entire role
```
mcieplak, line 192: Why don't we need to do this for the single role members as well? Little bit confused here as to why we need to differentiate, didn't the previous logic do the check for the role member?

kpharasi: So the check outside the loop is for a role member (which is basically a user/service/group). The expiry/system disabled options are either set for this role member or in case of a group it could also be set at the groupmember (all members in the group) level as well. The loop inside detects the scenario where the member is group member and then makes those additional checks. If it was a role member we have already made those checks outside and it would be repeating that. But, in case of groups there are two levels of checks that are required - 1. for the group present as a rolemember in the role and 2. for the particular group in question.

pkg/istio/rbac/common/util.go
kpharasi: True, will change

mcieplak (approved):

RuiGuo34 (approved):

PR #72, opened 2021-09-07
Adding logic for Combination Policies

kpharasi:
@mdhume @mcieplak @RuiGuo34 

### DIFF
--- a/pkg/athenz/model.go
+++ b/pkg/athenz/model.go
@@ -26,6 +26,8 @@ type RoleAssertions map[zms.ResourceName][]*zms.Assertion
 // map of Role:Members for an Athenz domain
 type RoleMembers map[zms.ResourceName][]*zms.RoleMember
 
+type RoleTags map[zms.ResourceName]map[zms.CompoundName]*zms.StringList
+
 // map of Group:GroupMembers for an Athenz domain
 type GroupMembers map[zms.MemberName][]*zms.GroupMember
 
@@ -37,6 +39,7 @@ type Model struct {
 	Rules        RoleAssertions `json:"rules,omitempty"`
 	Members      RoleMembers    `json:"members,omitempty"`
 	GroupMembers GroupMembers   `json:"groups,omitempty"`
+	RoleTags     RoleTags       `json:"roletags,omitempty"`
 }
 
 // getRolesForDomain returns the role names list in the same order as defined on the Athenz domain
@@ -113,6 +116,25 @@ func getMembersForRole(domain *zms.DomainData, crCache *cache.SharedIndexInforme
 	return roleMembers
 }
 
+// getTagsForRole returns the tags for each role in an Athenz domain
+func getTagsForRole(domain *zms.DomainData) RoleTags {
+	roleTags := make(RoleTags)
+
+	if domain == nil || domain.Roles == nil {
+		return roleTags
+	}
+
+	roles := domain.Roles
+	for _, role := range roles {
+		roleName := zms.ResourceName(role.Name)
+		if role.Tags != nil {
+			roleTags[roleName] = role.Tags
+		}
+	}
+
+	return roleTags
+}
+
 // getMembersForGroup returns the members for each group in an Athenz domain
 func getMembersForGroup(domain *zms.DomainData) GroupMembers {
 	// groupMembers creates a map where the key in the group name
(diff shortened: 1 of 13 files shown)


pkg/istio/rbac/v2/provider.go
```diff
@@ -253,13 +289,44 @@ func (p *v2) ConvertAthenzModelIntoIstioRbac(athenzModel athenz.Model, serviceNa
...
+		} else {
+			from_principal.Source.Principals = append(from_principal.Source.Principals, roleSpiffeName)
 		}
-		if p.enableOriginJwtSubject && len(from_requestPrincipal.Source.RequestPrincipals) > 0 {
-			rule.From = append(rule.From, from_requestPrincipal)
+
+		if combinationPolicyFlag {
+			rule.From = append(rule.From, from_principalAndRequestPrincipal)
+			rule.From = append(rule.From, from_principalAndNotRequestPrincipal)
+			if len(from_namespace.Source.Namespaces) > 0 {
```
mcieplak, line 319: Should we move the namespace check / append out of the if statements since it's the same for both?

pkg/istio/rbac/v2/provider.go
```diff
@@ -137,12 +151,21 @@ func (p *v2) ConvertAthenzModelIntoIstioRbac(athenzModel athenz.Model, serviceNa
...
 		}
 		from_requestPrincipal := &v1beta1.Rule_From{
 			Source: &v1beta1.Source{},
 		}
+
+		// Used if role has opted for combination policies
+		from_principalAndRequestPrincipal := &v1beta1.Rule_From{
+			Source: &v1beta1.Source{},
+		}
+		from_principalAndNotRequestPrincipal := &v1beta1.Rule_From{
```
mcieplak, line 166: I'm a little bit confused as to why we need this? If we have combination policies defined for an authz policy, do we not want it to be strictly be and AND statement of the proxy + token principal which is the from_principalAndRequestPrincipal?

mcieplak (changes requested):

pkg/athenz/model.go
```diff
@@ -113,6 +116,25 @@ func getMembersForRole(domain *zms.DomainData, crCache *cache.SharedIndexInforme
 	return roleMembers
 }
 
+// getTagsForRole returns the tags for each role in an Athenz domain
+func getTagsForRole(domain *zms.DomainData) RoleTags {
+	roleTags := make(RoleTags)
+
+	if domain == nil || domain.Roles == nil {
+		return roleTags
```
RuiGuo34, line 124: should we add debug log before return?

pkg/istio/rbac/v2/provider.go
kpharasi: This has been discussed.

pkg/istio/rbac/v2/provider.go
```diff
@@ -253,13 +289,36 @@ func (p *v2) ConvertAthenzModelIntoIstioRbac(athenzModel athenz.Model, serviceNa
...
-		rule.From = append(rule.From, from_principal)
-		if len(from_namespace.Source.Namespaces) > 0 {
-			rule.From = append(rule.From, from_namespace)
+		if combinationPolicyFlag {
+			from_principalAndRequestPrincipal.Source.Principals = append(from_principalAndRequestPrincipal.Source.Principals, roleSpiffeName)
+			from_principalAndNotRequestPrincipal.Source.Principals = append(from_principalAndNotRequestPrincipal.Source.Principals, roleSpiffeName)
+			for _, proxyPrincipal := range proxyPrincipalsList {
+				proxySpiffeName, err := common.MemberToSpiffe(proxyPrincipal)
+				if err != nil {
+					log.Errorln("error converting role member to spiffeName: ", err.Error())
```
mcieplak, line 298: This should be proxy principal instead of role member.

mcieplak (approved):

pkg/istio/onboarding/controller.go
```diff
@@ -5,11 +5,12 @@ package onboarding
...
 import (
 	"errors"
+	"time"
+
 	"github.com/yahoo/k8s-athenz-istio-auth/pkg/istio/rbac/common"
 	"istio.io/istio/pkg/config/schema/collections"
-	"time"
 
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
RuiGuo34, line 13: can just leave as v1

PR #73, opened 2022-11-16
Use Auth policy controller only

bhaskaramVaranasi:

## Goal
Instead of generating both alpha and the beta version of the controllers, only generate the authorization policy controllers.

### DIFF
--- a/main.go
+++ b/main.go
@@ -5,18 +5,23 @@ package main
 
 import (
 	"flag"
+	authzpolicy "github.com/yahoo/k8s-athenz-istio-auth/pkg/istio/authorizationpolicy"
+	adInformer "github.com/yahoo/k8s-athenz-syncer/pkg/client/informers/externalversions/athenz/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/client-go/tools/cache"
 	"os"
 	"os/signal"
 	"path/filepath"
 	"syscall"
 	"time"
 
-	"github.com/yahoo/k8s-athenz-istio-auth/pkg/controller"
 	"github.com/yahoo/k8s-athenz-istio-auth/pkg/istio/rbac/common"
 	"github.com/yahoo/k8s-athenz-istio-auth/pkg/log"
 	adClientset "github.com/yahoo/k8s-athenz-syncer/pkg/client/clientset/versioned"
 	versionedclient "istio.io/client-go/pkg/clientset/versioned"
 	crdController "istio.io/istio/pilot/pkg/config/kube/crd/controller"
+	istioController "istio.io/istio/pilot/pkg/serviceregistry/kube/controller"
 	"istio.io/istio/pkg/config/schema/collection"
 	"istio.io/istio/pkg/config/schema/collections"
 	"istio.io/pkg/ledger"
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -60,8 +67,12 @@ func main() {
 			log.Panicf("Error when creating authz policy directory: %s", err.Error())
 		}
 	}
-
-	configDescriptor := collection.SchemasFor(collections.IstioRbacV1Alpha1Serviceroles, collections.IstioRbacV1Alpha1Clusterrbacconfigs, collections.IstioRbacV1Alpha1Servicerolebindings, collections.IstioSecurityV1Beta1Authorizationpolicies)
+	var configDescriptor collection.Schemas
+	if *authPolicyControllerOnlyMode {
+		configDescriptor = collection.SchemasFor(collections.IstioRbacV1Alpha1Serviceroles, collections.IstioRbacV1Alpha1Clusterrbacconfigs, collections.IstioRbacV1Alpha1Servicerolebindings, collections.IstioSecurityV1Beta1Authorizationpolicies)
```
hk-vmg, line 72: Should L72 and L74 be the same?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -111,11 +113,16 @@ func (c *Controller) processEvent(fn cache.KeyFunc, obj interface{}) {
...
+	if c.standAloneMode {
+		go c.serviceIndexInformer.Run(stopCh)
+		go c.configStoreCache.Run(stopCh)
+		go c.adIndexInformer.Run(stopCh)
+	}
 
 	if !cache.WaitForCacheSync(stopCh, c.configStoreCache.HasSynced, c.serviceIndexInformer.HasSynced, c.adIndexInformer.HasSynced) {
 		log.Panicln("Timed out waiting for namespace cache to sync.")
 	}
+	go c.resync(stopCh)
```
hk-vmg, line 125: How does the behavior change if we resync after cache-sync instead of before?

pkg/istio/authorizationpolicy/controller.go
```diff
@@ -111,11 +113,16 @@ func (c *Controller) processEvent(fn cache.KeyFunc, obj interface{}) {
 
 // Run starts the main controller loop running sync at every poll interval.
 func (c *Controller) Run(stopCh <-chan struct{}) {
-	go c.resync(stopCh)
+	if c.standAloneMode {
```
hk-vmg, line 116: Can you help me understand `standAloneMode`? Is this equal to `authPolicyControllerOnly`?

main.go
bhaskaramVaranasi: No they are not supposed to be the same, thanks for catching that 👍🏽 . Updated the code to use only Beta version when running in the authPolicyControllerOnlyMode.

pkg/istio/authorizationpolicy/controller.go
bhaskaramVaranasi: In the previous version, the authorizationpolicy controller was instantiated by the controller (https://github.com/yahoo/k8s-athenz-istio-auth/blob/master/pkg/controller/controller.go#L154-L157)[here], the values like  `c.configStoreCache` , `c.serviceIndexInformer` and `c.adIndexInformer` are shared between these controllers. Before the `Run` from the autorizatioinpolicy controller was called, it waits for cache sync (https://github.com/yahoo/k8s-athenz-istio-auth/blob/master/pkg/controller/controller.go#L218-L227)[here]. 

As the we already waited on the cache sync in the controller already, the cache sync here is redundant for the authorization controller.

pkg/istio/authorizationpolicy/controller.go
bhaskaramVaranasi: The way I thought about it
`standAloneMode` - a setting for the authorizationpolicy controller to indicate that it needs to use the `Run` method of the  `serviceIndexInformer`, `configStoreCache ` and `adIndexInformer`, in the case where the value is false, these methods are run by the controller that instantiates the authorizationpolicy controller.

`authPolicyControllerOnly` - this is a command line parameter mainly used the in the `main` function to indicate that we would like to run the authpolicy controller only.

hk-vmg (approved):

